PR #242, opened 2014-06-13
Express: add index view support.

gillesfabio:
This PR :

* Move Express-related code to `src/express.js`
* Add Express index view support ([see](https://github.com/visionmedia/express/blob/master/lib/view.js#L62))
* Add `tests/express-sample/views/indexview` folder
* Add `tests/express-sample/views/indexview/index.html` file
* Add a new test in `tests/express/express.js`

Example.

Let's create a folder `views/blog` and `views/blog/index.html`.

Then to render the view: 

```js
app.get('/blog', function(req, res) {
  res.render('blog'); // same as 'blog/index' or 'blog/index.html'
});
```

It's the default Express behavior.

Tell me if you rather prefer to keep the Express-related code into `src/environment.js`. 

I'm not sure I correctly use environment's `loaders`. I just iterate over them and their `searchPaths` property to check the index file existence. It's maybe (probably?) not the right way.

### DIFF
--- a/src/environment.js
+++ b/src/environment.js
@@ -7,6 +7,7 @@ var builtin_filters = require('./filters');
 var builtin_loaders = require('./loaders');
 var runtime = require('./runtime');
 var globals = require('./globals');
+var express = require('./express');
 var Frame = runtime.Frame;
 
 var Environment = Obj.extend({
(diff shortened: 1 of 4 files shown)


jlongster:
@gillesfabio I'm so sorry I let this rot. I really like @garygreen's idea of having separate modules for various servers. This could totally be its own repo. Are you interested in pulling that out and maintaining it?

jlongster:
I'm looking over this again. This doesn't really work. You are iterating over all the loaders and assume they each have `searchPaths`, but that's just an internal property of `FileSystemLoader`. If you want to imitate this behavior, you need to do it in `FileSystemLoader`.

Express tries to take over control over how templates are resolved, and I don't like that. I wish it were more flexible. Nunjucks has its own mechanism for resolving templates, allowing you to have multiple loaders for searching several directories, or you could even write your own loader to load templates from a database. We also handle caching to make sure that all of this is available everywhere, including the browser.

Because of that, there's only so far our express integration can go. You are free to write your own integration layer though that provides the api described in http://expressjs.com/advanced/developing-template-engines.html. That way you'll delegate everything to express, like caching, and it will act exactly like you expect it to. It might actually be nice to have that module on npm, if you'd like to write it.

Issue #197, opened 2014-02-24
Else conditional not available inside For


david-duncan:
Given the example from Jinja docs:

``` jinja
<ul>
{% for user in users %}
    <li>{{ user.username|e }}</li>
{% else %}
    <li><em>no users found</em></li>
{% endfor %}
</ul>
```

One should be able to provide an else conditional inside of a for loop to easily render situations in which the stack is empty

Nunjucks 1.0.1
Uncaught Template render error: (quote-review.jinja) [Line 30, Column 10]
  unknown block tag: else

PR #259, opened 2014-07-22
Else clause in for loops

carljm:
This implements support for the `{% else %}` clause in a for loop, fixing #197.

I didn't implement support for the else clause in the async loops; since these loops don't exist in Jinja, there's no compatibility issue. If someone wants that support, I'm sure it's possible, I just don't need it and didn't feel like deciphering `_compileAsyncLoop`.


impactphoto:
This would indeed be very useful, and perhaps it should be considered as a bug given that nunjucks doesn't support breaking loops either (#296)?

PR #263, opened 2014-07-28
Added Filter typeof.

techmsi:


trss:
Helpful in debugging! Was stuck with a situation now where I pass a `string` or `null` and when `null` is passed the check `value === null` fails!!

However `{{ value | dump }}` prints "null" and the check `value == null` passes. So resorting to `==` in this situation alone for now.

revelt:
For posterity, the arguments laid out way higher are true but not apply in 100% of cases - data structures that template uses should, ideally, be known, stable and of the same type. 

I said _ideally_. I disagree with the way of thought that data structures should _always_ be fixed and of the same type. We should _strive_ to keep data types the same, but when it's necessary _multiple types of data_ simplifies code and makes life easier. Let me illustrate this.

For example, when templating, I need to wrap chunks of code with other code. Let's say you know it's always two elements `<code>` and `<alternative code>`, so "wrapping" content is object of 3 keys: heads, bellies and tails:

```
heads
<code>
bellies
<alternative code>
tails
``` 

wrapper's JSON:

```json
{
	"wrapper": {
		"heads": "some head wrapper",
		"bellies": "code that goes in between",
		"tails": "code that goes after"
	}
}
```

however, suddenly, different template comes and you are asked to wrap 5 chunks. For example, if-code-else-alternative-code PLUS extra IF-ELSE.

This means, "wrapping" content is array:

```json
{
	"wrapper": [
		"some head wrapper",
		"bellies, code that goes between 1st and 2nd chunks",
		"tails",
		"this new one",
		"and this one"
	]
}
```

**Now, existential question: do you recode all your templates to consume array instead of a plain object, or do you wire up your wrapping to wrap differently if contents are of an array-type?**

Recoding is not an option if that's production code and you can't just order full rounds of QA on everything that was already done ages ago, can you?

Mind you, array as wrapping structure appears superior - works for any amount of chunks, however, it also introduces a liability - what is order or mappings is incorrect? With object, you always could visually-`grep` that `tails` are `tails`. Now tails suddenly is `...[3]`. Are you sure it's not fourth element? And also, array as structure increases cognitive load since you start to forget what was that fifth value actually?

I'm going to solve this with custom filters in Gulp, so all is fine. Just wanted to leave mark here for posterity..

PR #277, opened 2014-09-05
Add syntax: {% include 'template' with {var: 2, var2: 'b'} %}

amyboyd:
This is useful for re-using bits of HTML that have small changes, for example a search box used on multiple pages with different placeholder text:

`{% include 'search-box.html.twig' with {placeholder: 'Search users'} %}`

`{% include 'search-box.html.twig' with {placeholder: 'Search companies'} %}`

This isn't a feature of jinja, but is a feature of the similar Twig framework.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,16 @@
+# EditorConfig helps developers define and maintain consistent
+# coding styles between different editors and IDEs
+# editorconfig.org
+
+root = true
+
+[*]
+indent_style = space
+indent_size = 4
+end_of_line = lf
+charset = utf-8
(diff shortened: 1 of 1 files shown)


Starefossen:
Too bad this didn't get accepted :disappointed:

dotmagic:
@jlongster I think most are aware the solution you show above is a possible workaround. But it does add another global variable and I have to use a second `set` to reset it after `include`. That's really not nice code anymore.

If we add the with parameter, nunjucks is still compatible with jinia, so why not this useful feature in core? Keep in mind Jinja started to implement the Template System from Django and was adding some new features too. ;)

PR #288, opened 2014-09-29
Set __filename and __dirname variables in templates.

amyboyd:


jlongster:
Sorry, not going to do this right now.

PR #302, opened 2014-10-18
Added option to customize the wrapper for precompiled templates

rhengles:
Users can use this options to store templates in modules, for example.

### DIFF
--- a/src/precompile-global.js
+++ b/src/precompile-global.js
@@ -1,7 +1,7 @@
 function precompileGlobal(name, template, opts) {
     opts = opts || {};
 
-		name = JSON.stringify(name);
+    name = JSON.stringify(name);
 
     var out = '(function() {' +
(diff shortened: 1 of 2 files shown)


rhengles:
Can you give me an idea of what you're considering refactor here? Can this be considered for 2.0 (#347)?

jlongster:
@rhengles I'm taking a look at this again, and I think we could do this with a minor bump (1.2 -> 1.3). I'd like to review this a little more before accepting it though.

jlongster:
Ok I'm definitely looking at this again. If we break the precompile API it may be reason enough for 2.0. No need to worry about bumping the major version.

jlongster:
You've convinced me, I'll merge this in and play with it locally to make sure everything works.

jlongster:
Actually, it looks like this is backwards-compatible. I thought you replaced the `asFunction` option but it's still there. this should be fine.

rhengles:
:+1: :100: Thanks man!

PR #304, opened 2014-10-20
Update `default` filter

cahnory:
Add the `boolean` argument documented in http://jinja.pocoo.org/docs/dev/templates/#default.

I wonder if `null` should be treated as undefined. (see https://github.com/mitsuhiko/jinja2/blob/master/jinja2/filters.py#L267)

### DIFF
--- a/src/filters.js
+++ b/src/filters.js
@@ -50,8 +50,8 @@ var filters = {
         return r.copySafeness(str, pre + str + post);
     },
 
-    'default': function(val, def) {
-        return val ? val : def;
+    'default': function(val, def, boolean) {
+        return val !== undefined && (!boolean || val !== false) ? val : def;
     },
 
     dictsort: function(val, case_sensitive, by) {


jlongster:
I think it's worth looking at this as a small backward incompatible change for 2.0. We could make `default` be strict and add another filter that does the same loose checking that it does now.

carljm:
I'm not sure that "another filter that does the same loose checking" is really worth it, given that `or` works equally well for that purpose and is actually less verbose.

jlongster:
That's true. If `or` is equivalent, we don't really need it.

jlongster:
@carljm thinking about this more, I'm really uncomfortable changing the current default filter. That's just one of those changes that might make it really hard to upgrade for some codebases (or worse, introduce subtle bugs).

I'm considering just introducing a new filter that is strict by default and just deprecating `default`, eventually we can remove it and tell people to use `or`.

carljm:
I don't think the upgrade would be that difficult (if you want the same semantics, just switch to `or` - it's doable with a regex search-and-replace). But it certainly could result in subtle bugs if people don't read the release notes and make the change.

From my perspective (where Jinja compatibility matters), adding a new filter under a different name with the Jinja behavior doesn't help at all. But if you want to do it, doesn't matter to me.

jlongster:
I've moved this over to https://github.com/mozilla/nunjucks/pull/428 where I am implementing the change in a safe way. People can use `a or b` to get the previous semantics, or pass true like `a | default(b, true)`

Issue #233, opened 2014-05-14
Asynchronous rendering throws compilation errors.


rcky:
Nunjucks throws compilation errors in the asynchronous API (see `environment.js`) instead of passing them to the callback.

PR #306, opened 2014-10-22
async rendering doesn't throw anymore when encountering compilation errors

eventEmitter:
fixes http://github.com/mozilla/nunjucks/issues/233

when calling the `render` or `renderString` method with a callback and the template contains a syntax error an error is thrown instead of returning the error to the callback. This patch fixes this behavior. Compilation errors are now returned to the callback.

### DIFF
--- a/src/environment.js
+++ b/src/environment.js
@@ -342,7 +342,13 @@ var Template = Obj.extend({
         }
 
         return lib.withPrettyErrors(this.path, this.env.dev, function() {
-            this.compile();
+            try {
+                this.compile();
+            } catch (e) {
+                if (cb) cb(e);
+                else throw e;
+                return;
+            }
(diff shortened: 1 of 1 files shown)


garygreen:
Thanks for this. Merged :-)

src/environment.js
```diff
@@ -342,7 +342,14 @@ var Template = Obj.extend({
         }
 
         return lib.withPrettyErrors(this.path, this.env.dev, function() {
-            this.compile();
+
+            // ctach compile errors for async rendering
```
popomore: typo

garygreen: Yup, spelling in another commit

popomore: Saw that :smile:

Issue #201, opened 2014-03-02
Infinite loop in replace filter


novaugust:
Replace will cause an infinite loop if the new string contains the old string, for example, if you're using replace to escape double-quotes like below.

    replace( '"', '\\"')   

Would there be any problem with just using the built-in javascript replace?

jlongster:
https://github.com/mozilla/nunjucks/pull/321

saml:
Can still reproduce in v1.3.4

Steps to reproduce:

```
$ cd /tmp
$ rm -rf node_modules/
$ npm install nunjucks
$ node
> require('nunjucks').renderString('{{ x|replace("a", "ab") }}', {x:'a'});
```

PR #321, opened 2014-11-14
Replace replace

dhendo:
Re-written to behave more like the Jinja2 implementation - it calls directly into python's replace().
- Now does not result in an infinite loop if no maxCount is used and the old string is a substring of replacement. fixes mozilla/nunjucks#201
- Handle numbers a bit better
- Improved tests
- Replacing an empty string will interleave the replacement


src/filters.js
```diff
@@ -191,19 +191,51 @@ var filters = {
...
+        if(typeof str === 'number'){
+            str = str +"";
+        }
+
+        if(typeof old !== "string") {
+            return str;
+        }
+
+        // ShortCircuits
+        if(!old){
```
jlongster: These `if` statements are confusing. Above you say if old is not a string, then return `str`, but here you are treating it like a boolean. If you mean to check for the empty string, please do a strict check like `old !== ''`. Can you combine the above `if` statements into `if/else` where that works too?

dhendo:
@jlongster Squished the ifs where possible. Updated empty string check and added some more tests.

luto:
@dhendo thank you for staying on this :)

dhendo:
Added more tests and a fix for when the input is not a string (e.g. undefined)

jlongster:
@dhendo so sorry it's taken so long to respond. can you address the test failures on travis? why is that failing?

dhendo:
It would have been helpful if I'd committed both files the first time round!

jlongster:
@dhendo yeesh I merged a bunch of other stuff and now it can't merge :/ can you rebase? sorry!

dhendo:
The rebase looks to have gone a bit awry - I'll re-apply on top of a clean branch and create a new PR.

dhendo:
https://github.com/mozilla/nunjucks/pull/420

PR #337, opened 2014-12-14
Node env support pre-compiled templates

ulion:
This PR based on https://github.com/mozilla/nunjucks/pull/302

by adopting the wrapper of compiled template, the compiled templates now support be loaded in Node env by the PrecompiledLoader.


jlongster:
Why do you want to precompile templates in node? They are cached so they are only compiled on the first load, and from then on always are loaded from the cached versions.

jlongster:
You mentioned something here: https://github.com/mozilla/nunjucks/pull/302#issuecomment-66867524 but can you elaborate? When can you not use the `fs` module?

ulion:
Because on some cloud env, they do not support use fs module to locate the
template file, e.g. leancloud, or parse.com (maybe)

2015-01-09 5:20 GMT+08:00 James Long <notifications@github.com>:

> Why do you want to precompile templates in node? They are cached so they
> are only compiled on the first load, and from then on always are loaded
> from the cached versions.
>
> —
> <https://github.com/mozilla/nunjucks/pull/337#issuecomment-69251244>.
>



-- 
Ulion

rhengles:
Hey @ulion, it's great that you used my PR! :D

I submitted a PR to your PR (Yo dawg!). In it you can clearly see the motivation for my PR - the ability to store precompiled templates in modules. Also, the browser doesn't get the checks for node env neither the other way around.

ulion:
@rhengles Did a great clean PR on my PR, I merged that and added cmd parameter for the bin/precompile command, now user can compile templates as module by bin/precompile -w cjs-module template-folder > template.js

ulion:
Oh, wait, the compiled template seems not directly work, I need adjust code still.

ulion:
OK, the code is good, just the code to include the compiled template need adjust:

before these change with this PR:
envOpts.precompiledTemplates = require('cloud/templates.js').nunjucksPrecompiled;

now with this PR:
envOpts.precompiledTemplates = require('cloud/templates.js');

jlongster:
@ulion sorry for letting this bitrot, I merged @rhengles' patch can you rebase?

ulion:
ok, will try to that now.

2015-04-28 6:22 GMT+08:00 James Long <notifications@github.com>:

> @ulion <https://github.com/ulion> sorry for letting this bitrot, I merged
> @rhengles <https://github.com/rhengles>' patch can you rebase?
>
> —
> <https://github.com/mozilla/nunjucks/pull/337#issuecomment-96837941>.
>



-- 
Ulion

ulion:
@jlongster rebased.

jlongster:
I'm still iffy about including all of this in nunjucks by default. Seems like a good candidate for a separate module, and I'm happy to add in whatever hooks needed to make that possible. I'll take a look at this again tomorrow though and give it a second though.

jlongster:
You can think of it like a webpack loader: the precompile script can take a "wrapper" that resolves to a node module that nunjucks calls out to. That way all of the CJS-specific stuff is in a separate module and you would just invoke it like `precompile --wrapper cjs` with the module `nunjucks-precompile-cjs` installed (it takes the wrapper argument and appends it to `nunjucks-precompile-`)

ulion:
browser and nodejs env is the 2 basic use env of the nunjucks, there seems
be no reason to only support one env precompiled templates. and there would
be less chance any other precompile module out there imo.

2015-04-28 8:56 GMT+08:00 James Long <notifications@github.com>:

> You can think of it like a webpack loader: the precompile script can take
> a "wrapper" that resolves to a node module that nunjucks calls out to. That
> way all of the CJS-specific stuff is in a separate module and you would
> just invoke it like precompile --wrapper cjs with the module
> nunjucks-precompile-cjs installed (it takes the wrapper argument and
> appends it to nunjucks-precompile-)
>
> —
> <https://github.com/mozilla/nunjucks/pull/337#issuecomment-96861426>.
>



-- 
Ulion

rhengles:
In the (near?) future, we'll need to add formatters for ES6 modules, and there's also AMD. I understand @jlongster 's opinion. It would be a more powerful abstraction and cleaner, too.

ulion:
ops, I have no idea about that.

If so, how do we implement required functions for what current PR does? I
have no idea.

2015-04-28 9:03 GMT+08:00 Rafael Hengles <notifications@github.com>:

> In the (near?) future, we'll need to add formatters for ES6 modules, and
> there's also AMD. I understand @jlongster <https://github.com/jlongster>
> 's opinion. It would be a more powerful abstraction and cleaner, too.
>
> —
> <https://github.com/mozilla/nunjucks/pull/337#issuecomment-96863586>.
>



-- 
Ulion

jlongster:
The reason for supporting browser precompilation out of the box is because it's not only the most common usage or precompilation, it's *required* for production use. Nunjucks has been around for years and this is the first request I've heard for other environments, so it's just a far less common case.

The precompile API exists so that you can do anything you want with it, and nunjucks doesn't have to worry about support more formats. I do see potential in being able to specify the format at the command line and it automatically invokes your custom formatter to make sharing formatters/loaders easy.

I'll whip up a patch to show you what I mean.

jlongster:
This is what I'd like to merge instead: #418

I went ahead and added in the `PrecompiledLoader` so that you don't always have to make that. Initially I thought that the plugin could also include a loader, but I think almost all precompiles will bundle modules in the same way.

Now, just create a new modules, say `nunjucks-cjs` and publish it to npm, and install it. Then you can just say `nunjucks-precompile -w cjs views` and it will `require` the nunjucks-cjs module and use it as the wrapper. The module must export a function as `wrapper`, like `module.exports.wrapper = ...`.

jlongster:
Let's move discussion over to https://github.com/mozilla/nunjucks/pull/418

Issue #326, opened 2014-11-24
let include use ../


jeremyBass:
in my case it would be very handy to be able to use `{% include "../pages/"+current_build+".html" -%}`

If I put the `pages` folder in the template folder and do `{% include "pages/"+current_build+".html" -%}` it works fine, but in this case the `pages` folder is context and shouldn't be in the template folder.  It would be nice if we could traverse the folders so I could keep my content and templates separate.

SamyPesse:
:+1:

devoidfury:
You can use multiple template folders by manually creating the loader instead of the automatic one.

```js
var cascade_loader = new nunjucks.FileSystemLoader(['./templates', './content']);
var env = new nunjucks.Environment(cascade_loader);
// ... hook env into your app or use it directly
```

popomore:
Agreed, it should lookup always inside the directory. Relative path is useful, and throw exception when out of the box.

SamyPesse:
@popomore :+1:

jlongster:
fixed in #349, it's a pretty simple fix actually and you can't reach "outside the box"

Issue #243, opened 2014-06-13
Absolute and Relative paths with Express


NoumanSaleem:
Trying to get nunjucks to work using Express and support both relative and absolute paths for layouts

```
/lib
 /views
  index.html
  layout.html
```
### index.html

```
{% extends "layout.html" %}
{% block content %}
{{ article.title }}
{% endblock %}
```
### app.js

``` javascript
  var env = new nunjucks.Environment(new nunjucks.FileSystemLoader(path.join(root, 'lib/views')));
  env.express(app);
```

when I change `extends "layout.html` to `extends layout`, with layout being `/Users/nsaleem/projects/app/lib/views/layout.html`
I get:

```
Error: template not found: /Users/nsaleem/projects/app/lib/views/layout.html at null.
```

thanks!

PR #349, opened 2015-01-12
Accept relative paths in import/include

### LINKED ISSUE #228: Inherit templates using relative paths
It would be very useful if, in addition to the template dir specified in the paths param in .configure, it would be super useful if templates could specify templates to inherit from by using relative paths.  

Any suggested workarounds?

SamyPesse:
This PR fixes #228,#326, #243. It simply lets you use `../` and `./` in **import** and **include**.

- [x] Pass path of current template to `env.getTemplate`
- [x] Add tests

### Example:

```
- folder1/
    - test.html
    - test2.html
- layout.html
```

test.html:
```html
{% extends ../layout.html %}

{% block test %}
Hello
{% endblock %}
```

test2.html:
```html
{% extends ./test.html %}

{% block test %}
Hello 2
{% endblock %}
```

In another matter, I don't think this line https://github.com/mozilla/nunjucks/blob/master/src/environment.js#L142 is a good idea, because if you're using an async loader inside your express application, the app crash if the loader return an error in the callback. The error should be passed to the callback (or throwned if sync): line 163.

### DIFF
--- a/src/compiler.js
+++ b/src/compiler.js
@@ -33,7 +33,8 @@ function quotedArray(arr) {
 }
 
 var Compiler = Object.extend({
-    init: function() {
+    init: function(name) {
+        this.name = name;
         this.codebuf = [];
         this.lastId = 0;
         this.buffer = null;
@@ -60,6 +61,7 @@ var Compiler = Object.extend({
     },
 
     emit: function(code) {
+        //console.log("emit", code);
         this.codebuf.push(code);
     },
 
@@ -891,7 +893,7 @@ var Compiler = Object.extend({
 
         this.emit('env.getTemplate(');
         this._compileExpression(node.template, frame);
-        this.emitLine(', ' + this.makeCallback(id));
+        this.emitLine(', false, "'+this.name+'", ' + this.makeCallback(id));
         this.addScopeLevel();
 
         this.emitLine(id + '.getExported(' +
@@ -914,7 +916,7 @@ var Compiler = Object.extend({
 
         this.emit('env.getTemplate(');
         this._compileExpression(node.template, frame);
-        this.emitLine(', ' + this.makeCallback(importedId));
+        this.emitLine(', false, "'+this.name+'", ' + this.makeCallback(importedId));
         this.addScopeLevel();
 
         this.emitLine(importedId + '.getExported(' +
@@ -989,7 +991,7 @@ var Compiler = Object.extend({
 
         this.emit('env.getTemplate(');
         this._compileExpression(node.template, frame);
-        this.emitLine(', true, ' + this.makeCallback('parentTemplate'));
+        this.emitLine(', true, "'+this.name+'", ' + this.makeCallback('parentTemplate'));
 
         this.emitLine('for(var ' + k + ' in parentTemplate.blocks) {');
         this.emitLine('context.addBlock(' + k +
@@ -1006,7 +1008,7 @@ var Compiler = Object.extend({
 
         this.emit('env.getTemplate(');
         this._compileExpression(node.template, frame);
-        this.emitLine(', ' + this.makeCallback(id));
+        this.emitLine(', false, "'+this.name+'", '+ this.makeCallback(id));
         this.addScopeLevel();
 
         this.emitLine(id + '.render(' +
--- a/src/compiler.js
+++ b/src/compiler.js
@@ -1106,7 +1106,7 @@ var Compiler = Object.extend({
 
 module.exports = {
     compile: function(src, asyncFilters, extensions, name, opts) {
-        var c = new Compiler();
+        var c = new Compiler(name);
 
         // Run the extension preprocessors against the source.
         if(extensions && extensions.length) {
--- a/src/environment.js
+++ b/src/environment.js
@@ -99,15 +99,15 @@ var Environment = Obj.extend({
         return this.filters[name];
     },
 
-    getTemplate: function(name, eagerCompile, fromName, cb) {
+    getTemplate: function(name, eagerCompile, parentName, cb) {
         if(name && name.raw) {
             // this fixes autoescape for templates referenced in symbols
             name = name.raw;
         }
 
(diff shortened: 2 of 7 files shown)


garygreen:
RE: line 142. I've been meaning to accept a PR for this #306, should fix this issue. Thanks for the PR. Does this work with async web loading the templates?

SamyPesse:
I just added tests for relative paths: https://github.com/mozilla/nunjucks/commit/4869a58dc139d96ee247aec10ba8bb38b9383d8c

Feel free to let me know if my tests are not following your guidelines.

I don't know for async web loading, is there any tests for it? The most of the logic is here: https://github.com/mozilla/nunjucks/pull/349/files#diff-b77b8f7ef4365a2872db79644add92f8R121
When `env.getTemplate` has a `parentName` argument, it resolve the current path to it using `path.resolve`.

One improvement could be to transform this `path.resolve` into a method from the loader (which is defaulting to `path.resolve`), so that loader can handle its own way the resolve of the path. I'll add this, then it'll be ready to be merged :rocket: 

I'm not an english native, so I'm sorry if my messages are not clear.

SamyPesse:
Ok, it should be good, I added a `resolve`method to the Loader object (https://github.com/SamyPesse/nunjucks/commit/927f2a6f5ec9b28ee2983ab146f469e557230683), so that Loader can resolve in their own way relative paths.

jlongster:
This is a much requested feature, so thanks for digging in and making a PR! Your approach looks good. I'm not convinced that we need to push the `resolve` method into loaders; paths are the definitive way to describe templates in the system, and nunjucks assumes that you define templates as paths (because it maps well for nested templates). Even if you loaded your templates from a db, for examples, you would still key off paths and relative resolution would still work normally. It looks like the way you did it, if you request `./foo.html` from `app/nested/bar.html`, it would resolve to the full path `app/nested/foo.html` and then get that template from the loader.

That approach seems fine to me and I would just hardcode that part of the resolution (at least for now). It would still work fine for loading templates over the web, as well as precompiled templates, because when it actually asks for the template it's using the full path.

SamyPesse:
Ok, I move the `resolve` into the loader because `path.resolve` can have some side effects (it returns an absolute url, ...) and I'm not sure if it will work if all use cases (ex: I don't know how browserify is handling `path.resolve`).
I can remove my last commit, just let me know.

SamyPesse:
I just added an option to the `env.renderString` method to accept an option to define the path to use for resolving paths. By default, this option is `null`.

I added a test for this.

SamyPesse:
Any news ? :)

fabien:
+1

SamyPesse:
Do you want me to rebase the PR or you want to close it?

jlongster:
Sorry, if you don't mind rebasing, that would be great!

SamyPesse:
Done!

jlongster:
Thanks! I'm pretty slammed at work right now but I'm definitely going to keep going through the PRs this week.

jlongster:
Neat, because it resolves the path at runtime, it seems to work pretty well. Precompiled templates still work normally, which I thought there might be problem with.

Thanks a lot for this! Sorry I didn't get around to looking through it until now. I'll release a new version soon.

SamyPesse:
Thanks :beers:

rhengles:
I am trying to use the PrecompiledLoader in the browser, but I'm getting this error:

TypeError: path.dirname is not a function
    at Obj.extend.resolve (http://localhost:8576/js/nunjucks-slim.js:2504:35)
    at Obj.extend.resolveTemplate (http://localhost:8576/js/nunjucks-slim.js:597:56)
    at Obj.extend.getTemplate (http://localhost:8576/js/nunjucks-slim.js:627:35)
    at root [as rootRenderFunc] (http://localhost:8576/js/<my-app-script>.js)
    at Obj.extend.render (http://localhost:8576/js/nunjucks-slim.js:923:16)
    at http://localhost:8576/js/nunjucks-slim.js:756:36
    at createTemplate (http://localhost:8576/js/nunjucks-slim.js:679:26)
    at handle (http://localhost:8576/js/nunjucks-slim.js:694:26)
    at http://localhost:8576/js/nunjucks-slim.js:708:22
    at next (http://localhost:8576/js/nunjucks-slim.js:347:14)

I found out it's because the precompiled templates still have the relative paths - they're not converted to absolute paths - and the nunjucks runtime is calling Loader#resolve() which uses "path.dirname" and "path.resolve" from require("path").

In my case, I don't know how, but the "path" variable has this value:

{
	"PrecompiledLoader": function () {
	        if(prototype.init) {
	            prototype.init.apply(this, arguments);
	        }
	}
}

How do I solve this? I'm using nunjucks 2.4.2.

Thanks!

rdebeasi:
Any plans to release this feature in a future version? Thanks! :)

carljm:
It's already in multiple released versions.

rdebeasi:
Whoops - looks like I was using an npm package that pulled in an outdated version of nunjucks. Sorry for the confusion!

electerious:
Does this still work? I can't find a test for `../` and 

```
{% extends ../layout.html %}

{% block test %}
Hello
{% endblock %}
```

isn't working for me. It says

```
Template render error: (/Users/tobiasreich/Sites/Templates/Skeleton NJK/src/assets/components/button/button.njk)
 Error: template not found: /Users/tobiasreich/Sites/Templates/Skeleton NJK/src/assets/components/layout/layout.njk
```

but the template exists. FileSystemLoader searchPath points to `/Users/tobiasreich/Sites/Templates/Skeleton NJK/src/assets/components/button/`.

electerious:
It works when I add `/Users/tobiasreich/Sites/Templates/Skeleton NJK/src/assets/components/` to the searchPath of the FileSystemLoader. Is this the correct solution?

electerious:
My code seems to fail because `../layout/layout.njk` is "outside the box". Is there anything I can do to allow extends/imports "outside the box"?

PR #354, opened 2015-01-15
Add option to disable cache in FileSystemLoader

SamyPesse:


SamyPesse:
I made this PR because when developing a web application, you want to test the pages without caching the templates. The watcher option slows everything (at least for me).

jlongster:
I'm down with this, but how is the watcher slower?

Soon we're going to make the watcher piece more modular, letting you choose which watcher to use. Right now we are bound to chokidar which I've found tends to have various issues. Maybe another watcher is better?

Still, this sounds reasonable. We are at the point in this project where most changes are just more configuration options :)

PR #355, opened 2015-01-17
Implement lstripBlocks and trimBlocks from Jinja2

radev:
This pull request implements functionality from Jinja2 trim_blocks and lstrip_blocks (renaming to camelCase).
It can significantly reduce amount of whitespace in the output if a lot of block tags are used.

### DIFF
--- a/src/compiler.js
+++ b/src/compiler.js
@@ -1103,7 +1103,7 @@ var Compiler = Object.extend({
 // console.log(tmpl);
 
 module.exports = {
-    compile: function(src, asyncFilters, extensions, name, lexerTags) {
+    compile: function(src, asyncFilters, extensions, name, lexerTags, trimBlocks, lstripBlocks) {
         var c = new Compiler();
 
         // Run the extension preprocessors against the source.
@@ -1115,7 +1115,11 @@ module.exports = {
             }
         }
 
-        c.compile(transformer.transform(parser.parse(src, extensions, lexerTags),
+        c.compile(transformer.transform(parser.parse(src,
+                                                     extensions,
+                                                     lexerTags,
+                                                     trimBlocks,
+                                                     lstripBlocks),
                                         asyncFilters,
                                         name));
         return c.getCode();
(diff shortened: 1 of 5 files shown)


garygreen:
Thanks for this. Only briefly looked at the code, does this handle windows-style line breaks too \r\n?

src/compiler.js
```diff
@@ -1103,7 +1103,7 @@ var Compiler = Object.extend({
 // console.log(tmpl);
 
 module.exports = {
-    compile: function(src, asyncFilters, extensions, name, lexerTags) {
+    compile: function(src, asyncFilters, extensions, name, lexerTags, trimBlocks, lstripBlocks) {
```
garygreen: I think we should rethink this, the number of parameters being passed here is a lot.

radev:
No, this doesn't support windows-style line breaks, but Jinja2 doesn't support them too:
https://github.com/mitsuhiko/jinja2/blob/c97fa109d38b5367ce3efffb6e435d007dc55920/jinja2/lexer.py#L434

src/compiler.js
radev: We can pass env instead of lexerTags, trimBlocks, lstripBlocks

radev: So, should I pass environment to lexer?

jlongster: Let's pass around a `options` object instead. If you can, move all the arguments after `src` into an `options` argument and us it like `opts.trimBlocks`. I think it would be fine to pass this same options argument to the parser and lexer as well.

jlongster:
Thanks for doing this, it looks good.

src/compiler.js
radev: Problem is that path is not from opts and it's fourth argument. So either asyncFilters and extensions should be left where they are now. Or we break backwards compatibility.

garygreen: Don't care about breaking backwards compatibility as this will be going in nunjucks 2.0

radev:
I've moved some Environment options to opts, but left filters, asyncFilters, extensions as is &mdash; they are mutable by Environment.

jlongster:
Thanks, I like how this looks. I'd say we go ahead and merge it and if we really want to support windows we can do that in the future.

magnusjt:
Will you be supporting windows-style line breaks? This issue had me confused for quite some time..

carljm:
@magnusjt I'm not aware of anyone currently planning to implement support for Windows-style line breaks. I'd be happy to look at a pull request.

Issue #189, opened 2014-02-16
most of string filters (capitalize, center, indent, truncate...) failed if left value is not a string


tdelmas:
The comportment of nunjucks is very permissive in some ways : 

"If a value is undefined or null, nothing is displayed. The same behavior occurs when referencing undefined or null objects. The following all output nothing if foo is undefined: {{ foo }}, {{ foo.bar }}, {{ foo.bar.baz }}."

But the simple fact to add a filter can break a template if the value is undefined : 
If foo is undefined, 
{{ foo }} display nothing but {{ foo ¦ truncate }} failed.

I think most string filters for strings should start with :
if ( typeof str !== 'string' ) { return ''; }

metasansana:
Uhh +1... What's the status of this? I was about to create a custom `capitalize` tag and came across this issue. Can I help?

garygreen:
I don't think nunjucks should automatically convert `undefined` or `null` to `''` before pipe to the filter, as some filters can handle arrays, objects, etc. Though I agree it should definitely be handled by the filter itself, to check for instances of an undefined/null and convert to the most 'sane empty' type for that filter so it won't throw an exception.

metasansana:
@garygreen I can do a PR for what @tdelmas suggests for capitalize, center, indent, truncate. I don't see it doing any harm in the interim.

jlongster:
Let's move this into the PR here: https://github.com/mozilla/nunjucks/pull/358 We'll fix it there.

jlongster:
This is fixed on master

PR #358, opened 2015-01-24
Add tests and workaround for #189

metasansana:


garygreen:
Thanks for this, there's a few issues though e.g. if you pass through a non-string, like a number, it should still handle it.

I have an idea for a more generic implementation which I will try and get around to later

shainab:
@garygreen  Ok, Let me know if you want me to modify this instead then.

metasansana:
@garygreen Ok, Let me know if you want me to modify this instead then.

jlongster:
Yeah. This is basically silently failing if the value isn't a string. I suspect most of the time the value *should* be a string, and I personally like being a little more strict about that. The behavior right now errors, basically telling the user that they did something wrong. However, I'm fine if we go the other way with some of these builtin filters and coerce the value to a string instead.

We definitely don't want to show nothing if it's not a string. That seems ripe for confusion. ("why isn't this showing up??" and then they have to spend time figuring out they accidentally passed an integer)

We could follow the same rules as we do when outputting a value: only show nothing for `null` & `undefined` (and `false`? I can't remember). Otherwise, we just call `toString()` on it.

garygreen:
Well I was thinking about a basic map of 'default values' if your passing a value that is either `null` or `undefined` to a filter. Could even make it an option like `strict_filters` or something. Was thinking of something along the lines of the below, where it would check the emptyDefaults mapping before passing the value onto the filter if it's null or undefined:

```javascript
var emptyDefaults = {
	abs: 0,
	center: '',
	first: [],
	join: []
};
```

jlongster:
That's a good point; each filter wants to default to something different. I think it might be more appropriate as a utility function though, which each filter calls with the value (like `normalize(val, [])`) and gives it a default.

garygreen:
Sounds good

metasansana:
There is a `default` filter. Should I call that then pass it to `string`? Or maybe add `normalize` and a `stringify` method to `lib`?

jlongster:
I would say do the following:

1. Create just a simple top-level function in `filters.js` that looks like this:

```js
function value(v, default) {
  if(v === null || v === undefined || v === false) {
    return default;
  }
  return v;
}
```

2. Use this at the top of the filters that should support falsy values. So at the top of  `center`, for example, it would be `str = value(str, '')`.

That way there's just one place where we specify the rules of how to treat values.

garygreen:
I would not do the `v === false` check as that assumes too much, this is more to do with nullable/undefined values.

jlongster:
I'm down with that.

jlongster:
This was fixed in #403, we basically did the same thing we talked about here for normalizing the inputs.

metasansana:
I fell into a black hole with some client work, sorry. :(  Thank you all for addressing this though.

PR #364, opened 2015-02-04
Allow list filter to pass arrays through as-is.

carljm:
It's a bit silly for the `|list` filter to say "type not iterable" when given an array; it should just pass the array through unchanged instead.

This PR does that, and also adds a test for the currently-untested case of passing an object to `|list`.


jlongster:
Absolutely.

PR #365, opened 2015-02-04
Add support for selectattr and rejectattr filters.

carljm:
The `select` and `reject` filters from Jinja rely on the concept of "tests", which nunjucks doesn't have. But `selectattr` and `rejectattr` are quite useful, and the single-argument version of them does not rely on tests.

### DIFF
--- a/src/filters.js
+++ b/src/filters.js
@@ -186,6 +186,18 @@ var filters = {
         return arr[Math.floor(Math.random() * arr.length)];
     },
 
+    rejectattr: function(arr, attr) {
+      return arr.filter(function (item) {
+        return !item[attr];
+      });
+    },
+
+    selectattr: function(arr, attr) {
+      return arr.filter(function (item) {
+        return !!item[attr];
+      });
+    },
+
     replace: function(str, old, new_, maxCount) {
         if (old instanceof RegExp) {
             return str.replace(old, new_);
--- a/tests/filters.js
+++ b/tests/filters.js
@@ -210,6 +210,18 @@
             finish(done);
         });
 
+        it('rejectattr', function(done) {
+          var foods = [{tasty: true}, {tasty: false}, {tasty: true}];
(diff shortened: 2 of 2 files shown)


jlongster:
I'm down with this. I think the names are a little weird, but it's fine with me. Feel free to merge!

jlongster:
Eh I'll go ahead and merge it :) I really need to merge the docs into this repo, because I'd force you to write some docs! Going to do that soon.

carljm:
Ha, yes :-) I'll leave that as motivation for making the docs merge happen.

jlongster:
Hey @carljm I just did it. Can you add docs here? https://github.com/mozilla/nunjucks/blob/master/docs/templating.md#builtin-filters

carljm:
Sure; done in 3dd3158 - thanks for porting the docs over!

jameswragg:
Really missing the tests you mention here. Use case:
`{{ users|selectattr("email", "equalto", "foo@bar.invalid") }}`

How can I help?

devoidfury:
@jameswragg the source for those two is here, if you'd like to take a try at it and submit a PR: https://github.com/mozilla/nunjucks/blob/master/src/filters.js#L243

Here is the list of tests - http://jinja.pocoo.org/docs/dev/templates/#list-of-builtin-tests -- I don't believe we have anything like that currently, so those would need to be implemented as well.

Alternatively, you could write a custom filter to hook into your own code -- something like this:

```js
env.addFilter('findEqual', function(arr, attr, match) {
	if (!Array.isArray(arr)) return
	return arr.filter(function(item) {
		return item && item[attr] == match
	})
})
```

```jinja2
{{ users|findEqual("email", "foo@bar.invalid") }}
```

jameswragg:
@devoidfury that's great, thank you! Filters seem pretty simple, thanks for the intro.

PR #370, opened 2015-02-09
emit update when the template is changed the first time when the event =...

MaxKramnik:
...= "add". The template would not update when it is modified first time. The watcher listens only to the "change" event. The first event is actually "add".

### DIFF
--- a/src/node-loaders.js
+++ b/src/node-loaders.js
@@ -26,10 +26,10 @@ var FileSystemLoader = Loader.extend({
             // they change
             lib.each(this.searchPaths, function(p) {
                 if(existsSync(p)) {
-                    var watcher = chokidar.watch(p, { ignoreInitial: true });
+                    var watcher = chokidar.watch(p);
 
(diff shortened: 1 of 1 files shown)


jlongster:
Isn't `add` when a new file is created? Are you saying that when you first fire up the app, load a page, and change an existing template the updates don't show up the first time? I can't reproduce that.

MaxKramnik:
James,

Yeah, load the page, then make a template file change, refresh the page, the change is not there. If you make another change, save, refresh the page, both changes show up.

Thanks,

Max

jlongster:
I think we can take the `ignoreInitial` option off actually. Without it the `add` event is emitted immediately, and we will just ignore it. The first save should then be a `change` event. I can't remember why we added that option. Chokidar may have changed behavior in a recent version.

Can you update your PR to remove that option?

MaxKramnik:
Removing ignoreInitial worked, thanks!

Issue #378, opened 2015-02-21
Support for relative paths introduces cache performance bug


boutell:
See #349 cc: @SamyPesse 

With the new support for relative paths in folder names, the same "name" can refer to two completely different templates.

The cache storage code does in fact address this, storing the cached template under its resolved name.

However the cache fetch code runs before the resolve call.

The correct resolution would be to put the resolve call before the cache fetch code. However there is a wrinkle: nunjucks permits multiple loaders at once. And the "resolve" call currently is not responsible for saying "this is the correct resolution *and* I'm sure this file actually exists and I'm taking responsibility for it, so stop looking for other loaders."

I think the "resolve" API needs to be tried for each configured loader, and must become responsible for not returning a happy result unless the file at the resolved path actually exists.

SamyPesse:
I'm taking a look at the issue, I'll add some unit tests and propose a fix.

boutell:
I'm working on unit tests now, if I can get my install --dev to complete...

On Sat, Feb 21, 2015 at 11:01 AM, Samy Pessé <notifications@github.com>
wrote:

> I'm taking a look at the issue, I'll add some unit tests and propose a fix.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/mozilla/nunjucks/issues/378#issuecomment-75377866>.
>



-- 


*THOMAS BOUTELL, *DEV & OPS
P'UNK AVENUE | (215) 755-1330  |  punkave.com

boutell:
Hmm, my test passed. Is the cache always active?

On Sat, Feb 21, 2015 at 11:04 AM, Tom Boutell <tom@punkave.com> wrote:

> I'm working on unit tests now, if I can get my install --dev to complete...
>
> On Sat, Feb 21, 2015 at 11:01 AM, Samy Pessé <notifications@github.com>
> wrote:
>
>> I'm taking a look at the issue, I'll add some unit tests and propose a
>> fix.
>>
>> —
>> Reply to this email directly or view it on GitHub
>> <https://github.com/mozilla/nunjucks/issues/378#issuecomment-75377866>.
>>
>
>
>
> --
>
>
> *THOMAS BOUTELL, *DEV & OPS
> P'UNK AVENUE | (215) 755-1330  |  punkave.com
>



-- 


*THOMAS BOUTELL, *DEV & OPS
P'UNK AVENUE | (215) 755-1330  |  punkave.com

SamyPesse:
[...]

PR #379, opened 2015-02-21
Improve performance and reliability of relative paths implementation

### LINKED ISSUE #375: Current API imposes a Unix-filesystem-like convention on relative path interpretation
I know you don't support relative paths in templates by default (at least, not last I looked), but I'm aware that nunjucks is *very* extensible, to the point of being able to add tags and so on. So I thought you'd be up for hearing about the issues I ran into when I chose to implement this myself.

Template A wants to include template B which lives in another folder, so it uses a relative path, with a custom loader that permits this. All is good so far.

Now template B wants to include template C. It just uses the name of the file, because template C is in the same folder.

Here we have problems (:

A nunjucks loader is only passed the name being included. It is not passed any information about the "parent" path the file is being loaded *from*.

In the case of Apostrophe, I'm not dealing with relative paths exactly: I'm dealing with "cross-module includes," where instead of "../../whatever.html" I want to include "modulename:whatever.html". So my loader resolves what folder "modulename" corresponds to. But it amounts to the same thing: given the current API, it (almost) isn't possible to support the use of non-absolute paths in template B when it attempts to load template C.

But if the loader received "parentName," the name under which template B was loaded, then it would be trivially possible to do any sort of relative path resolution we wish.

OK, I say "almost isn't possible" because I did come up with a hack. But it is a terrible hack, and you're going to see it, and wince, and wish that loaders received parentName. Here it is:

```javascript
  // Fix all calls to extend, include and import so that
  // paths without an explicit module name use the specified
  // module name. Uses the official Nunjucks lexer. -Tom

  self.resolveTemplatePathsViaModule = function(src, moduleName) {
    var out = '';
    var lexer = templates.nunjucks.lexer.lex(src);
    var token;
    var state = 'boring';
    while (token = lexer.nextToken()) {
      // Push the token back out again. Unfortunately
      // there is no .raw property, so we have to quote
      // strings and regexes again ourselves (and hope
      // there's nothing else weird out there...)
      if (token.type === 'string') {
        out += '"' + token.value.replace(/\"/g, '\\\"') + '"';
      } else if (token.type === 'regex') {
        out += '/' + token.value.replace(/\//g, '\\/') + '/';
      } else {
        out += token.value;
      }
      if (token.type === 'block-start') {
        state = 'block';
      } else if ((token.type === 'symbol') && (_.contains([ 'include', 'import', 'extends', 'from' ], token.value))) {
        state = 'expr';
        // This is naive and can introduce additional module names
        // if the path here in the template already has a module name.
        // This is worked around explicitly in getSource(). -Tom
        out += ' "' + moduleName + ':" +';
        state = 'boring';
      } else if (token.type === 'block-end') {
        state = 'boring';
      }
    }

    return out;
  };
```

Wow, isn't that awful? I use your tokenizer to modify the include, import and from statements in the template being included, prepending a module path to them.

This totally works, and could also be used to support relative paths in a filesystem loader that allows multiple directories of templates. But I'm forced to rebuild the source as I go. And because there is no ".rawValue" I have to rebuild quoted strings and regexps... and pray those are the only tokens where ".value" and the original raw code would be different. Yikes. (:

I could ask for .rawValue in the lexer, and maybe I should because that would be useful to someone, but I think supplying parentName to loaders would be a lot more practical.

How do you feel about that idea? Would you accept such a pull request?

Thanks!

SamyPesse:
This PR fix #375, #381 and #378.


SamyPesse:
Any news? :)

jlongster:
Looking now!

src/compiler.js
```diff
@@ -121,6 +121,10 @@ var Compiler = Object.extend({
         return 't_' + this.lastId;
     },
 
+    _templateName: function() {
+        return this.templateName == null? "undefined" : '"'+this.templateName.replace(/"/g, '\\"')+'"';
```
jlongster: When would templateName have quotes in it/

src/environment.js
```diff
@@ -147,8 +152,7 @@ var Environment = Obj.extend({
                 }
 
                 // Resolve name relative to parentName
-                if (parentName && (name.indexOf("./") == 0 ||
-                                   name.indexOf("../") == 0)) {
```
jlongster: How come you removed the check for a relative path? We don't want to resolve every single template. Wouldn't that break all current templates? A normal template like `foo.html` should resolve to the root folder. I don't see any other changes related to this.

src/environment.js
```diff
@@ -120,7 +119,11 @@ var Environment = Obj.extend({
             throw new Error('template names must be a string: ' + name);
         }
 
-        var tmpl = this.cache[name];
+        for (var i = 0; i < this.loaders.length; i++) {
+            var _name = parentName? this.loaders[i].resolve(parentName, name) : name;
+            tmpl = this.loaders[i].cache[_name];
```
jlongster: What does moving the cache into the loaders fix again? Since you are running the resolution here as well, doesn't that fix the caching problem?

src/environment.js
jlongster: I know there is some discussion about this in other issues, but I still don't understand. Won't this break all templates that extend other ones (or includes)? Any includes in that template will now be relative by default instead of absolute.

SamyPesse: Yes, I'm looking at this, so adding a `isRelative` method to the loader is a best idea (to avoid this: https://github.com/mozilla/nunjucks/issues/375), will that be good?

src/environment.js
SamyPesse: This is related to https://github.com/mozilla/nunjucks/issues/378#issuecomment-75380107

src/environment.js
jlongster: That makes sense to me. I like that better than forcing all loaders to implement this same logic if they want to copy the default behavior. If you want to do what @boutell wants, you just always return `true` from that method.

src/environment.js
jlongster: That's true. I'm not sure I love that we are going down a route of allowing almost anything to be possible (sometimes constraints make it easier to reason about any nunjucks project), but I think this project generally is evolving that way, so I'm fine with that. At this point I'm pretty much open to whatever you all need.

src/environment.js
SamyPesse: Take a look at 4568ace8cfadbf3af9f954e1004c6acc16468863 and let me know what you think :)

src/compiler.js
SamyPesse: Filename can contains quotes (at least on OS X).

I can remove this, if you believe this is not necessary.

src/environment.js
```diff
@@ -99,7 +97,14 @@ var Environment = Obj.extend({
         return this.filters[name];
     },
 
+    resolveTemplate: function(loader, parentName, filename) {
+        var isRelative = (loader.isRelative && parentName)? loader.isRelative(filename) : false;
+        return (isRelative && loader.resolve)? loader.resolve(parentName, filename) : filename;
```
jlongster: Looks good! One last nit: I think we can avoid checking is the methods exist and just call `isRelative` and `resolve` since they are on the base loader, and it this point we expect loaders to inherit from it.

So just `parentName ? loader.isRelative(filename) : false`, etc

src/compiler.js
jlongster: Ok. It doesn't hurt. Eventually we could have a top-level method in `compiler.js` that does this quote escaping, but this is fine for now.

src/environment.js
SamyPesse: This is related to #381, let me know if I should ignore the issue or keep the exist check.

jlongster: Ah, ok, that's fine. The evolution of a codebase. :p

jlongster:
I think this is a good to go, right?

SamyPesse:
I think this is ready, but it's your decision ;)

SamyPesse:
Let me know if you need me to do anymore changes :) and thanks again for all your work on Nunjucks :+1

boutell:
Yes, although I have a loader that just implements the interface and does
not inherit, I agree that means I'm breaking the rules and it's up to me to
keep that working if I want to do that (:

On Fri, Mar 6, 2015 at 12:44 PM, Samy Pessé <notifications@github.com>
wrote:

> Let me know if you need me to do anymore changes :) and thanks again for
> all your work on Nunjucks :+1
>
> —
> <https://github.com/mozilla/nunjucks/pull/379#issuecomment-77602012>.
>



-- 


*THOMAS BOUTELL, *DEV & OPS
P'UNK AVENUE | (215) 755-1330  |  punkave.com

jlongster:
Alright, we can support legacy loaders for now. At some point in the future (maybe 2.0) it might be nice to force users to upgrade to the Loader base class. It doesn't seem like it would be that hard, but we can  deal with that in the future.

SamyPesse:
Thanks for merging :+1: :beers: When will the new version be published to npm?

jlongster:
Thanks for all the hard work! I can make a new version later next week. I'd like to let it sit for a few days at least.

SamyPesse:
Ok perfect :) thanks, we still have a few things to finish before releasing GitBook 2.0.0 as stable and we prefer to use Nunjucks from NPM: https://github.com/GitbookIO/gitbook/blob/version/2.0/package.json#L16

PR #380, opened 2015-03-03
add keywords in package.json

stevemao:
I was trying to search for "templating system" but I couldn't find this module. I think at least these two words should be added.


garygreen:
Thanks

PR #383, opened 2015-03-06
Improved code style

### LINKED ISSUE #1: "use strict";
Use the strict mode of javascript on all the files: "use strict";

popomore:
- [x] use strict mode #386
- [x] add jshint for code lint
- [x] use === and !== instead == and != #387
- [x] remove unuse variable
- [x] always use single quote #385
- [x] travis support: remove node@0.8, add iojs

Closes #1

### DIFF
--- a/src/environment.js
+++ b/src/environment.js
@@ -3,13 +3,13 @@
 var path = require('path');
 var lib = require('./lib');
 var Obj = require('./object');
-var lexer = require('./lexer');
 var compiler = require('./compiler');
 var builtin_filters = require('./filters');
 var builtin_loaders = require('./loaders');
 var runtime = require('./runtime');
 var globals = require('./globals');
 var Frame = runtime.Frame;
+var Template;
 
 var Environment = Obj.extend({
     init: function(loaders, opts) {
@@ -19,7 +19,7 @@ var Environment = Obj.extend({
         // (the full trace from within nunjucks may confuse developers using
         //  the library)
         // defaults to false
-        var opts = this.opts = opts || {};
+        opts = this.opts = opts || {};
         this.opts.dev = !!opts.dev;
 
         // The autoescape flag sets global autoescaping. If true,
@@ -331,7 +331,7 @@ var Context = Obj.extend({
     }
 });
 
-var Template = Obj.extend({
+Template = Obj.extend({
     init: function (src, env, path, eagerCompile) {
         this.env = env || new Environment();
 
@@ -447,6 +447,7 @@ var Template = Obj.extend({
                                           this.path,
                                           this.env.opts);
 
+            /* jslint evil: true */
             var func = new Function(source);
             props = func();
         }
--- a/src/nodes.js
+++ b/src/nodes.js
@@ -2,6 +2,7 @@
 
 var util = require('util');
 var lib = require('./lib');
+// jshint -W079
 var Object = require('./object');
 
 function traverseAndCheck(obj, type, results) {
@@ -20,7 +21,7 @@ var Node = Object.extend('Node', {
         this.colno = colno;
 
         var fields = this.fields;
-        for(var i=0, l=fields.length; i<l; i++) {
+        for(var i = 0, l = fields.length; i < l; i++) {
             var field = fields[i];
 
             // The first two args are line/col numbers, so offset by 2
@@ -39,17 +40,18 @@ var Node = Object.extend('Node', {
     findAll: function(type, results) {
         results = results || [];
 
+        var i, l;
         if(this instanceof NodeList) {
             var children = this.children;
 
-            for(var i=0, l=children.length; i<l; i++) {
+            for(i = 0, l = children.length; i < l; i++) {
                 traverseAndCheck(children[i], type, results);
             }
         }
         else {
             var fields = this.fields;
 
-            for(var i=0, l=fields.length; i<l; i++) {
+            for(i = 0, l = fields.length; i < l; i++) {
                 traverseAndCheck(this[fields[i]], type, results);
             }
         }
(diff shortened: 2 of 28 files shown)


src/loaders.js
```diff
@@ -1,6 +1,10 @@
-if(typeof window === 'undefined' || window !== this) {
-    module.exports = require('./node-loaders');
-}
-else {
-    module.exports = require('./web-loaders');
-}
+'use strict';
+
+(function() {
+    if(typeof window === 'undefined' || window !== this) {
```
popomore: http://jslinterrors.com/option-validthis-cant-be-used-in-a-global-scope

src/nodes.js
```diff
@@ -104,41 +108,42 @@ var FromImport = Node.extend("FromImport", {
...
+var Mod = BinOp.extend('Mod');
+var Pow = BinOp.extend('Pow');
+var Neg = UnaryOp.extend('Neg');
+var Pos = UnaryOp.extend('Pos');
+var Compare = Node.extend('Compare', { fields: ['expr', 'ops'] });
+var CompareOperand = Node.extend('CompareOperand', {
     fields: ['expr', 'type']
 });
 
-var CustomTag = Node.extend("CustomTag", {
```
popomore: I don't know why this isn't used?

src/precompile.js
```diff
@@ -56,25 +56,7 @@ function precompile(input, opts) {
     else if(pathStats.isDirectory()) {
         var templates = [];
 
-        function addTemplates(dir) {
```
popomore: It's not good that define `function declaration` in `if` block

src/loaders.js
jlongster: The comparison doesn't make sense anymore, you've just made `this` always undefined in strict mode. I would ignore this file for linting.

src/nodes.js
jlongster: It's used be users, when they create custom extensions.

src/precompile.js
jlongster: Why? It's a more functional style and I think it's perfectly fine if abstracting out that logic makes sense. I wouldn't worry about refactoring this file because I'm already refactoring it with a new API (there's an issue but I forget the number)

jlongster:
Thanks for cleaning up, but please break this up into multiple PRs! There's no way I feel safe merging in something this huge. Can you first make a PR that safely makes all quotes single-quotes? And then follow-up with more smaller PRs?

src/nodes.js
popomore: But it's not exported?

src/loaders.js
popomore: That's all right, I ignore this file.

popomore:
What you think whether add jshint first or not? If I add jshint first, the minor change will break it.

Otherwise, I take PR first, then add jshint after all of PRs are merged.

src/precompile.js
popomore: http://jslinterrors.com/function-statements-should-not-be-placed-in-blocks

The original code will not break, but it's not good in ES5 because of function scope.

src/nodes.js
jlongster: Ah, you are correct, it must be leftover from the initial custom tags work. `CallExtension` is the node that this one became. We can safely remove it.

jlongster:
Do it whichever order you think makes sense, I'm fine either way. It sounds like it's easier to add jshint last after we've fixed up some of the code.

popomore:
@jlongster I have create three split PR, How does it look to you? Any suggestion?

popomore:
I'll do the rest in this PR, I think diff will more less than before.

jlongster:
What's the status of this? Are you going to rebase it?

popomore:
Yes, I'll rebase master.

popomore:
It's ready to be reviewed. Sorry for late push, I'm busy these days.

popomore:
@jlongster do you have time to review this and make a release.

src/environment.js
```diff
@@ -3,15 +3,15 @@
...
 var compiler = require('./compiler');
 var builtin_filters = require('./filters');
 var builtin_loaders = require('./loaders');
 var runtime = require('./runtime');
 var globals = require('./globals');
 var Frame = runtime.Frame;
+var Template, Environment;
 
-var Environment = Obj.extend({
+Environment = Obj.extend({
```
jlongster: I'm sorry, but I don't like these changes. Does it really complain about a `var` here? Can you make it so that it doesn't?

src/environment.js
```diff
@@ -19,7 +19,7 @@ var Environment = Obj.extend({
         // (the full trace from within nunjucks may confuse developers using
         //  the library)
         // defaults to false
-        var opts = this.opts = opts || {};
+        opts = this.opts = opts || {};
```
jlongster: Why are all of these `var`s being dropped?  I don't see them declared anywhere else. I don't like rules that force you declare variables completely separately from where they are used.

jlongster:
@popomore can you address the above comments? I see a few places where you dropped `var` that doesn't make much sense to me

src/environment.js
popomore: Because `opts` is parameter defined in function, so you should not declear it.

src/environment.js
popomore: Roll back.

popomore:
Pushed again, rebased master and fix minor problem.

popomore:
Test has passed!

jlongster:
Ok, cool thanks. I'll merge this. Seems a little scary, but it's mostly lots of nits here and there. Good job!

PR #384, opened 2015-03-07
Add contribute.json file.

Osmose:
Add a contribute.json file a la http://contribute.paas.allizom.org/.

### DIFF
--- a/contribute.json
+++ b/contribute.json
@@ -1,0 +1,22 @@
+{
+    "name": "Nunjucks",
+    "description": "Nunjucks is a full featured templating engine for javascript. It is heavily inspired by jinja2.",
+    "repository": {
+        "url": "https://github.com/mozilla/nunjucks",
+        "license": "BSD 2-Clause",
+        "tests": "https://travis-ci.org/mozilla/nunjucks"
+    },
+    "participate": {
+        "home": "https://github.com/mozilla/nunjucks",
+        "docs": "https://github.com/mozilla/nunjucks"
+    },
+    "bugs": {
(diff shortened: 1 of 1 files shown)


Osmose:
@jlongster r?

garygreen:
I'm really not sure we need this. Doesn't `package.json` have pretty much exactly this information?

Osmose:
> I'm not sure we need this. Doesn't package.json have pretty much exactly this information?

Main things missing are the participate section (which Nunjucks doesn't really have anything special for, which is why I just put the Github homepage for them), buglist, and tests parts. In my particular case, I'm working on https://github.com/mozilla/webdev which is hoping to use those for special buttons in the project listing like a "How can I help?" button for projects that have a participate section or listing issues from the buglist.

Also, since contribute.json is language-agnostic, I don't want to end up making a bunch of exceptions for languages/ecosystems that already have things that sort've come close to the info in contribute.json. Downside is that there will be some duplicated information, but it's worth trying IMO.

jlongster:
Hm, in the past I have fought hard to keep all the extraneous files at the root of the project to a minimum. I think it would be worthwhile for the webdev project to extract information from `package.json` if it detects it exists. You could probably do something similar for Python projects. You could still support a `contribute.json` files for projects that want to specifically control it, but if you re-used existing files (that really do already have all this info) you'd get a ton of projects for free already without having to create this.

jlongster:
As this is hosted under mozilla, and we need more contributors anyway, let's go ahead do this. We can remove it if the webdev page can ever get enough info from package.json. fwiw here's what uses it: http://mozilla.github.io/webdev/

PR #385, opened 2015-03-07
Change always using single quote

popomore:
Split PR from #383


jlongster:
Thanks. I don't like touching this many files at once (it pollutes the commit history for stuff like `git blame`), but I do prefer to use single-quotes and consistency is nice. Hopefully merging this doesn't cause too many problem with the other PRs.

PR #386, opened 2015-03-07
Use strict

popomore:
Split PR from #383 

Based on #385, merge this PR after that.

### DIFF
--- a/tests/globals.js
+++ b/tests/globals.js
@@ -1,5 +1,7 @@
 (function() {
-    var expect, util, lib;
+    'use strict';
+
+    var expect, util, lib, Environment, Loader, templatesPath;
 
     if(typeof require != 'undefined') {
         expect = require('expect.js');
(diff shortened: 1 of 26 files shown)


jlongster:
I merged #385 but for some reason this PR isn't updating to show that master already has the single quote commit... Do you know what it hasn't updated? I can't do the code review until I see the single commit for this.

popomore:
I don't know why, I try to rebase for it.

tests/globals.js
```diff
@@ -1,5 +1,7 @@
 (function() {
-    var expect, util, lib;
+    'use strict';
+
+    var expect, util, lib, Environment, Loader, templatesPath;
```
popomore: Should define a variable before use it in strict mode.

popomore:
It can be reviewed now.

jlongster:
Thanks. All tests are passing, so that's good. Let's hope it doesn't break anything else.

PR #387, opened 2015-03-07
use === and !== instead == and !=

popomore:
Split PR from #383

Based on #385 and #386, merge this PR after that.

### DIFF
--- a/src/compiler.js
+++ b/src/compiler.js
@@ -240,7 +240,7 @@ var Compiler = Object.extend({
                 // object as the last argument, if they exist.
                 this._compileExpression(arg, frame);
 
-                if(i != args.children.length - 1 || contentArgs.length) {
+                if(i !== args.children.length - 1 || contentArgs.length) {
                     this.emit(',');
                 }
             }, this);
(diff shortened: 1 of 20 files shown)


jlongster:
Same thing happened here, it's showing 3 commits for some reason.

popomore:
It has been rebased.

src/compiler.js
```diff
@@ -511,7 +511,7 @@ var Compiler = Object.extend({
             var name = target.value;
             var id = frame.lookup(name);
 
-            if (id == null) {
+            if (id === null || id === undefined) {
```
jlongster: I'm actually open to still using `==` for checking null/undefined. It gets annoying to always have to check for both, and once you learn that `==` matches both, I don't think its too confusing. That's the only case I'd want to use `==`. What do you think (others who see this, feel free to comment too)?

popomore: The practice leads us to always use `===` that `==` will confuse other what is exactly equal to.

But it works fine in this case.

jlongster:
Thanks!

PR #389, opened 2015-03-10
Fix error when template is not found

SamyPesse:
This is a fix for an error I introduced in my precedent PR (Sorry ...).

I'm currently running some tests for GitBook on Windows and Nunjucks is not working there (tests are all failing).
You should maybe consider using Appveyor (as simple as travis and free for open source projects) to test Nunjucks on Windows.
I'll take a deep look at why Nunjucks is not working (it's reason why GitBook is not working).

### DIFF
--- a/src/environment.js
+++ b/src/environment.js
@@ -149,9 +149,8 @@ var Environment = Obj.extend({
 
             lib.asyncIter(this.loaders, function(loader, i, next, done) {
                 function handle(src) {
-                    src.loader = loader;
-
                     if(src) {
+                        src.loader = loader;
                         done(src);
                     }
                     else {


jlongster:
Interesting, I know there are other windows users that use nunjucks. Maybe something recent broke it. I'll take a look at Appveyor, thanks.

SamyPesse:
@jlongster it looks like the problem come from my PRs for relative paths, so it's probably my fault.

I'm looking at it, but it will be great if you can setup a project on Appveyor associated with nunjucks so that in the future it avoids breaking Windows compatibility.

SamyPesse:
I think the problem is coming from here: https://github.com/mozilla/nunjucks/blob/master/src/compiler.js#L127

Basically windows path with `\` like `C:\test\hello` are not escaping so the paths become invalid.

I'm working on a fix and will propose a PR in the coming hours.

PR #390, opened 2015-03-10
fix watching in node-loader

jlongster:
@SamyPesse The relative templates broke watching. In `getSource` the `fullpath` key in `pathsToNames` is now always fully absolute, but in the watcher the path it gets is still just a relative path. I think this fixes it, but can you verify?

Are you not using the watcher?

### DIFF
--- a/src/node-loaders.js
+++ b/src/node-loaders.js
@@ -31,6 +31,7 @@ var FileSystemLoader = Loader.extend({
                     var watcher = chokidar.watch(p, { ignoreInitial: true });
 
                     watcher.on('all', function(event, fullname) {
+                        fullname = path.resolve(fullname);
                         if(event === 'change' && fullname in this.pathsToNames) {
                             this.emit('update', this.pathsToNames[fullname]);
                         }


src/node-loaders.js
```diff
@@ -31,6 +31,7 @@ var FileSystemLoader = Loader.extend({
                     var watcher = chokidar.watch(p, { ignoreInitial: true });
 
                     watcher.on('all', function(event, fullname) {
+                        fullname = path.join(process.cwd(), fullname);
```
SamyPesse: Why not `path.resolve` ? It's using `resolve` in `getSource`

jlongster: That works, I forgot about that.

SamyPesse:
I'm using the watcher but also the noCache option that's maybe why I didn't see this bug.

src/node-loaders.js
SamyPesse: Why not `path.resolve(p, fullname)` ?

jlongster:
Yeah, `noCache` would still work because this is a bug in cache invalidation.

src/node-loaders.js
```diff
@@ -31,6 +31,7 @@ var FileSystemLoader = Loader.extend({
                     var watcher = chokidar.watch(p, { ignoreInitial: true });
 
                     watcher.on('all', function(event, fullname) {
+                        fullname = path.resolve(fullname);
```
SamyPesse: Is `fullname` (from chokidar events) relative to `p` or absolute ?

Because if the path is relative, this will work only if `p == process.cwd()` which is not always the case.

jlongster: `fullname` is the path that changed according to the path you gave chokidar, `p` in this case. If `p` is absolute, `fullname` will be absolute (I think). If it's relative, it will be relative. If `p` is `views`, `fullname` will be `views/index.html`.

So basically this should work because chokidar's watch patch is relative to `process.cwd()`. If it isn't, you gave it an absolute path.

SamyPesse: ok :)

PR #391, opened 2015-03-10
Fix nunjucks for windows: relative paths and tests

SamyPesse:
The goals of this PR are:

- [x] Fix relative paths on Windows
- [x] Adapt tests for windows (some tests are using `\n` and on windows it's `\r\n`)
- [x] Maybe if you want to, adapt tests script to work with Appveyor

### DIFF
--- a/src/compiler.js
+++ b/src/compiler.js
@@ -124,7 +124,7 @@ var Compiler = Object.extend({
     },
 
     _templateName: function() {
-        return this.templateName == null? 'undefined' : '"'+this.templateName.replace(/"/g, '\\"')+'"';
+        return this.templateName == null? 'undefined' : JSON.stringify(this.templateName);
     },
 
     _bufferAppend: function(func) {
--- a/tests/util.js
+++ b/tests/util.js
@@ -54,7 +54,7 @@
 
     function normEOL(str) {
         if (!str) return str;
-        return str.replace(/\r\n|\r|\n/g, "\n");
+        return str.replace(/\r\n|\r/g, "\n");
     }
 
     function render(str, ctx, opts, cb) {
--- a/tests/util.js
+++ b/tests/util.js
@@ -54,7 +54,7 @@
 
     function normEOL(str) {
         if (!str) return str;
-        return str.replace("\r\n", "\n");
+        return str.replace(/\r\n|\r|\n/g, "\n");
     }
 
     function render(str, ctx, opts, cb) {
--- a/tests/util.js
+++ b/tests/util.js
@@ -52,6 +52,11 @@
         }
     }
 
+    function normEOL(str) {
+        if (!str) return str;
+        return str.replace("\r\n", "\n");
+    }
+
     function render(str, ctx, opts, cb) {
         if(!opts) {
             cb = ctx;
@@ -95,7 +100,7 @@
                     throw err;
                 }
 
-                cb(err, res);
+                cb(err, normEOL(res));
 
                 doneAsyncs++;
 
@@ -110,12 +115,14 @@
         module.exports.render = render;
         module.exports.equal = equal;
         module.exports.finish = finish;
+        module.exports.normEOL = normEOL;
     }
     else {
         window.util = {
             render: render,
             equal: equal,
-            finish: finish
+            finish: finish,
+            normEOL: normEOL
         };
     }
 })();
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,0 @@
-
-test:
-	node_modules/.bin/istanbul cover node_modules/.bin/_mocha \
-		-- -b -R tap tests
-
-browserfiles:
-	./bin/bundle browser/nunjucks
-	SLIM=1 ./bin/bundle browser/nunjucks-slim
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -7,7 +7,6 @@ environment:
   matrix:
     - nodejs_version: "0.11"
     - nodejs_version: "0.10"
-    - nodejs_version: "0.8"
 
 # Install scripts. (runs after repo cloning)
 install:
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,28 @@
+# Fix line endings in Windows. (runs before repo cloning)
+init:
+  - git config --global core.autocrlf input
+
+# Test against these versions of Node.js.
+environment:
+  matrix:
+    - nodejs_version: "0.11"
+    - nodejs_version: "0.10"
+    - nodejs_version: "0.8"
+
+# Install scripts. (runs after repo cloning)
+install:
+  # Get the latest stable version of Node.js or io.js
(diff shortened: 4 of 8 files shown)


src/compiler.js
```diff
@@ -124,7 +124,7 @@ var Compiler = Object.extend({
     },
 
     _templateName: function() {
-        return this.templateName == null? 'undefined' : '"'+this.templateName.replace(/"/g, '\\"')+'"';
+        return this.templateName == null? 'undefined' : JSON.stringify(this.templateName);
```
SamyPesse: This probably not looks like perfect but this escape correctly `\`, `"`, ...

SamyPesse:
The main problem with unit tests is that we are comparing strings that include `\n` like at https://github.com/mozilla/nunjucks/blob/master/tests/compiler.js#L326

This is failing on Windows because it's replaced by `\r\n`.

@jlongster Should I adapt the tests to use `os.EOL` or should we normalize the output or `render` to normalize EOLs.

jlongster:
@SamyPesse hm that's tricky. Any chance you know what other projects do? Is it safe to normalize it?

SamyPesse:
This is maybe safer to just normalize tests for now, for example at https://github.com/mozilla/nunjucks/blob/master/tests/compiler.js#L326:

```
expect(normEOL(res)).to.be('somecontenthere\n');
```

with:

```
function normEOL(s) {
    return s.replace("\r\n", "\n");
}
```

ok ?

jlongster:
That works for me!

jlongster:
I'd say we only need to do it in tests where we need to compare the `\n`.

SamyPesse:
Tests are now working on Windows, I'm testing setup tests on Appveyor (for my fork).

the use of `make` for tests is not really usefull and I'm not usre it'll work by default on windows (make is not installed with node).
The use of `make` at https://github.com/mozilla/nunjucks/blob/master/package.json#L23 can be replaced by only using npm scripts. Can I include this in my PR ?

SamyPesse:
BTW here are the builds that I'm testing using AppVeyor: https://ci.appveyor.com/project/SamyPesse/nunjucks

SamyPesse:
There is a very important issue in Mocha for windows that has been fixed in a merged PR but not yet release to NPM: https://github.com/mochajs/mocha/issues/333#issuecomment-51425925

jlongster:
What the issue we experience because of that bug? They say they won't backport to 0.10 but I still want to support that.

SamyPesse:
Basically stdout and stderr are not flushed and so not visible in the build logs. The only other workaround is to use the `--no-exit` option of mocha, but in this case builds fail because of timeout (but logs are available ...).

I'm running some tests, but I fix this is not important if logs are not available for some windows builds, it's enough to know that it's failing and getting the logs for node 0.11, because if it's failing only for 0.10/0.8, it will probably fail for the same reason on travis (where the logs are always available).

Sorry if my comment is not clear enough, I'm not an english native :)

SamyPesse:
:tada: :balloon: :rocket: Tests are passing on AppVeyor: https://ci.appveyor.com/project/SamyPesse/nunjucks/build/job/94ovau3gea1co8nr !

At least on 0.11, I'm still waiting for 0.8 and 0.10: https://ci.appveyor.com/project/SamyPesse/nunjucks

SamyPesse:
@jlongster this is ready to be merged :)

I disabled windows builds for node 0.8 because it was always failing without being related to Nunjucks: https://ci.appveyor.com/project/SamyPesse/nunjucks/build/1.0.3/job/nd0nj203yqeods0c
Having tests for 0.10 and 0.11 on Windows is enough and better than nothing.

You can signup to appveyor and create a project for nunjucks if you want to enable windows testing, otherwise at least it's ready for it :)

Nijikokun:
:+1:

tests/util.js
```diff
@@ -52,6 +52,11 @@
         }
     }
 
+    function normEOL(str) {
+        if (!str) return str;
+        return str.replace(/\r\n|\r|\n/g, "\n");
```
popomore: `\n` needn't be replaced, right?

SamyPesse: You mean using `/\r\n|\r|/g` instead? I think it will work fine, yes.

src/compiler.js
popomore: Why use `JSON.stringify` here?

tests/util.js
popomore: Yes

src/compiler.js
SamyPesse: Because we need to escape the `templateName` to be work in a javascript string, I think it's most secure way to do it, we need to handle `\`, `"`, `'`, ...
By using `.replace`, we may forget one case or not handle the combinaison.

What is you suggestion?

tests/util.js
SamyPesse: Ok I'll replace it

src/compiler.js
SamyPesse: One other solution could be the use of https://github.com/joliss/js-string-escape

popomore: Sound good.

SamyPesse: Which solution?

popomore:
LGTM

src/compiler.js
popomore: I think `JSON.stringify` is ok in this case that is only replace `"` originally.

SamyPesse:
Thanks :+1: :beers:

jlongster:
@SamyPesse thanks for your work on it! As far as Appveyor is concerned, is your project here (https://ci.appveyor.com/project/SamyPesse/nunjucks) testing mozilla's project or your local fork? I added Mozilla's repo but its showing up under my name: https://ci.appveyor.com/project/jlongster/nunjucks

We can keep it under your name if its testing this repo. I'll add a link to the README.

SamyPesse:
It's testing my fork, I can't enable it to test this repo (since I don't have access to the repo webhooks).

popomore:
https://ci.appveyor.com/project/jlongster/nunjucks should work now.

@jlongster you can register a mozilla account if want the link `project/mozilla/nunjucks`, or rename your account.

http://help.appveyor.com/discussions/questions/744-using-appveyor-for-a-github-org

jlongster:
Ok, thanks. I'll keep it under jlongster because I don't want to manage the mozilla account at Appveyor. As long as it's testing this repo.

SamyPesse:
Any news on when the latest updates will be available on NPM? :)

jlongster:
Because of some of the bugs we found I wanted to make sure there wasn't anything else outstanding with this. And I got busy recently. I will put some time into it this week though and make a release if it looks stable.

jlongster:
@SamyPesse just pushed out v1.3.0. Thanks for all your work! Is there any chance you could add documentation for the features you added in the /docs folder? Particularly for relative templates, but also for the `noCache` option in `FileSystemLoader`.

PR #393, opened 2015-03-17
Add option to automatically indent blocks

poke:
Since I care a lot about how the rendered markup of template looks, I have been trying to find a way to ensure that the indentation matches the original when extending a base template. This pull requests adds an option `indentBlocks` that does exactly that. As it compiles the templates, it will remember the indentation of blocks and automatically figure out how to indent the content of the blocks in extending templates to match the original indentation.

Since this might be undesired by some (not sure why though ;P), and since it actually broke a test (a *single* test—but only because of a trailing line ending that got removed), I made this option not a default. So people are free to decide whether they want this or not.

I’ve added tests that not only check the basic functionality but also ensure that it works when extending multiple levels of templates, or when definining a new block in an extending template that is then filled in a third level. The general rule is that the indentation of the block where it was originally defined decides about how the content is indented. For nested blocks, this is additive; for example in one test, there is a block `C` defined within the content of block `B` and the content `C` is indented for both `B` and `C`.

Here is a more simple example that shows how it works:

    <div>
      {% block A %}{% endblock %}
      <div>
         {% block B %}
         default content
         {% endblock %}
      </div>
    </div>

Example extending template:

    {% extends 'base.html' %}
    {% block A %}
    <h1>Example content</h1>
    {% endblock %}

    {% block B %}
    This content will have the original indentation.
    {% endblock %}

Above results in the following, which is nice and pretty:

    <div>
      <h1>Example content</h1>
      <div>
        This content will have the original indentation.
      </div>
    </div>

As for the implementation, I had to add the `formatBlocks` function somewhere where it was accessible from the compiled template code. I thought it was most fitting in the `runtime` module, at least compared to the other persistent alternatives (environment and context). For remembering the indentation, I used the `frame` object which luckily was not only persistent for the duration of the template execution but also was set up in a good order.

I originally tried to do the remembering and calculations directly in the compiler but unfortunately there wasn’t a way that allowed me to store the indentation state without making it completely global. This is because the `frame` object is only temporary for each block. So in the end, I had to put the logic into the compiled template code.

Any way, I think this turned out pretty nicely and it works pretty well for me, so I hope you will accept this change.

### DIFF
--- a/src/compiler.js
+++ b/src/compiler.js
@@ -954,12 +954,17 @@ var Compiler = Object.extend({
     },
 
     compileBlock: function(node, frame) {
-        if(!this.isChild) {
+        if(this.isChild) {
+            this.emitLine('if (!frame.hasOwnProperty("indent_' + node.name.value + '")) { frame["indent_' + node.name.value + '"] = ' + node.colno + '; }');
+        }
+        else {
             var id = this.tmpid();
 
             this.emitLine('context.getBlock("' + node.name.value + '")' +
                           '(env, context, frame, runtime, ' + this.makeCallback(id));
-            this.emitLine(this.buffer + ' += ' + id + ';');
+            this.emitLine('var indent = ' + node.colno + ' - (frame["indent_' + node.name.value + '"] || 1);');
+            this.emitLine('if (!frame.hasOwnProperty("indent_' + node.name.value + '")) { indent = 0; }');
+            this.emitLine(this.buffer + ' += runtime.formatBlock(' + id + ', indent, !env.opts.indentBlocks);');
             this.addScopeLevel();
         }
     },
--- a/src/compiler.js
+++ b/src/compiler.js
@@ -1006,6 +1006,7 @@ var Compiler = Object.extend({
     },
 
     compileInclude: function(node, frame) {
+        // console.log(node.colno);
         var id = this.tmpid();
         var id2 = this.tmpid();
 
@@ -1016,7 +1017,8 @@ var Compiler = Object.extend({
 
         this.emitLine(id + '.render(' +
                       'context.getVariables(), frame.push(), ' + this.makeCallback(id2));
-        this.emitLine(this.buffer + ' += ' + id2);
+        this.emitLine(this.buffer + ' += runtime.formatBlock(' + id2 + ', ' + (node.colno - 1) + ', !env.opts.indentBlocks);');
+
         this.addScopeLevel();
     },
 
--- a/src/environment.js
+++ b/src/environment.js
@@ -32,6 +32,8 @@ var Environment = Obj.extend({
 
         this.opts.lstripBlocks = !!opts.lstripBlocks;
 
+        this.opts.indentBlocks = !!opts.indentBlocks;
+
         if(!loaders) {
             // The filesystem loader is only available client-side
             if(builtin_loaders.FileSystemLoader) {
--- a/src/runtime.js
+++ b/src/runtime.js
@@ -320,6 +320,17 @@ function asyncAll(arr, dimen, func, cb) {
     }
 }
 
+function formatBlock(text, indentWidth, doNothing) {
+    if (doNothing) {
+        return text;
+    }
+    function indentLines(line, idx) {
+        return idx == 0 ? line : indent + line;
+    }
+    var indent = lib.repeat(' ', indentWidth);
+    return text.trim().split('\n').map(indentLines).join('\n');
+}
+
 module.exports = {
     Frame: Frame,
     makeMacro: makeMacro,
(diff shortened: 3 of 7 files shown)


jbergstroem:
Couldn't this be done by passing html to tidy or similar at a later stage? Just feels out of scope for a template engine. Worst case, you can still achieve this by using a combination of indents and `{%- spaceless stuff -%}`. I guess my point is that the goal of a template engine is different from one of style guides.

Also (moving into philosophy, which I guess is similar to your goal) since Safari, Chrome and Firefox nowadays gives you a rendered, formatted tree - the visual impact is pretty low.

poke:
I personally think the `{-` and `-}` always results in a lot of trial and error to make sure that whitespace is correctly removed (and most of the time it won’t work the way you want it to work). And it won’t do anything about indentation.

As for being able to pass things to tidy, sure that works, but I actually think that a template engine should be able to produce the desired output without having to pass it through another tool that then works based on its own rules without caring about the original formatting. And since this change just enforces whatever indentation was originally used for blocks, you can even take this further and use it now to render whitespace-sensitive output. For example you could generate Python code with it (or even do stupid things like render HAML with it). Keep in mind that template engines are not necessarily just about generating HTML output; they can be used for any kind of text.

As for the “formatted tree” in browser dev tools, these only show the current DOM tree but not really the original uninterpreted input, so it’s not the same.

jbergstroem:
@poke, thanks for your quick reply.

> `{-` and `-}` always results in a lot of trial and error

I agree. It's non-trivial and requires the user to keep track of indentation.

> Keep in mind that template engines are not necessarily just about generating HTML output; they can be used for any kind of text.

Agreed. This still doesn't mean that the template engine should auto-indent for you. In my opinion, a template engine should be extremely limited in scope -- not care too much about it's semantics.

When you refer to Python, yaml/haml or even jade (which are indent-sensitive) I would argue that it doesn't imply that it's underlying editor/parser automatically indents for you.

> As for the “formatted tree” in browser dev tools, these only show the current DOM tree but not really the original uninterpreted input, so it’s not the same.

You're correct. My point is that what you're looking at is the preferred (browser) view of the source, not necessarily the actual html source. On the other hand – if the preferred view is an improved representation, that's one point scored for the user.

Finishing it off, I don't have strong opinions against this commit - I just prefer having more arguments on the table. @poke -- thanks for the contribution!

poke:
> This still doesn't mean that the template engine should auto-indent for you.

It isn’t exactly auto-indentation as in that it magically figures out how the indentation is supposed to be. It just keeps whatever indentation was originally used for the blocks and keeps that base indentation for content that is rendered in the block. So it’s more about staying consistent to the base template ;)

> When you refer to Python, yaml/haml or even jade (which are indent-sensitive) I would argue that it doesn't imply that it's underlying editor/parser automatically indents for you.

What I meant is that you could for example generate valid Python code with this functionality. For example you could have a class with some `additional_methods` block inside and in an extending template you could just define those methods in the block without having to know at what indentation level the class members would need to be. The template just inserts them correctly for you. So even if the Python example is somewhat off for a JavaScript templating engine (I will eventually backport this functionality to Jinja2 though), it’s still a possible use case.

> Finishing it off, I don't have strong opinions against this commit - I just prefer having more arguments on the table.

Since this is purely optional, I don’t think it’s a bad thing to have in a templating engine, especially since the change didn’t turn out that complicated. And as for me, I have been wanting this functionality in template engines forever; and I know others that wouldn’t mind having more control over the indentation either.

poke:
The new commit also adds the same functionality to include tags.

popomore:
I think this PR is a fix rather than a feature, it rendered template as I want.

jlongster:
The problem is that even if a feature is by default off, any new code in the engine adds complexity, enlarges the surfaces for bugs, and adds maintenance costs.

This doesn't seem like it adds much code, but it's touching the core part of the compiler so I'm really hesitant to add it in. I think I need to see other people's vote on this to be worth it, sorry. I don't think most people care about the output like @jbergstroem said because you usually use the devtools to inspect it anyway.

Small papercuts in the runtime code also ends up slowing down an engine, too. Thanks a lot for your contribution, I will reconsider this if others chime in and express a desire for this.

poke:
Thank you for your response. I understand that you don’t want to increase the complexity of the library unnecessarily although I personally don’t think that this patch is very complex if you put it in perspective with the rest of the library.

As I’ve mentioned above, I’ll eventually submit an equivalent patch to jinja2. Would you reconsider your decision if it landed there?

jlongster:
@poke sure, if it lands there it helps your cause. if others in here simply vote yes too that helps. It's doesn't add *much* but that code is already hard enough to grok, and it's just another papercut for making it really hard to understand everything going on, and the return just isn't enough for me (in my opinion, you are abusing the `frame` data structure).

PR #403, opened 2015-04-08
Makes length filter work with SafeString

pierreant-p:
The `length` filter does not work with `SafeString` because `SafeString` is an `Object`, not a `string` or an `array`.
This pull request adds a type check in the length filter, so that in the case of a SafeString object, the length of the underlying value is returned.

It would probably be cleaner to add a `length` property to the `SafeString` object (like it was in version 1.0.2 and previous), but since the `SafeString` object inherits from `String.prototype`, the `length` property is read-only. I couldn't find a clean way to work around this limitation, so I decided change the `length` filter directly instead.

I'm happy to help if someone has a better idea to solve this issue.

### DIFF
--- a/src/filters.js
+++ b/src/filters.js
@@ -150,8 +150,12 @@ var filters = {
         return arr[arr.length-1];
     },
 
-    length: function(arr) {
-        return arr !== undefined ? arr.length : 0;
+    length: function(val) {
+        var value = val;
+        if ( val instanceof r.SafeString )
+            value = val.valueOf();
+
+        return value !== undefined ? value.length : 0;
     },
 
     list: function(val) {
--- a/src/filters.js
+++ b/src/filters.js
@@ -3,6 +3,21 @@
 var lib = require('./lib');
 var r = require('./runtime');
 
+
+var normalize = function(value, defaultValue) {
+
+    if (value === null || value === undefined) {
+        return defaultValue;
+    }
+
+    if (value instanceof r.SafeString) {
+        return normalize(value.valueOf(), defaultValue);
+    }
+
+    return value;
+};
+
+
 var filters = {
     abs: function(n) {
         return Math.abs(n);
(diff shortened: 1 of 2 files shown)


garygreen:
This could tie in with #358 as I suspect there are other filters that would benefit from this too. It would be good to have the `normalize

pierreant-p:
Updates the PR with the changes suggested in #358 so this should solve both issues.
Tests added. I took the liberty to copy some of @metasansana  's tests.

jlongster:
Yeah, `SafeString` is kind of weird. It implements `valueOf` so it should work in a lot of places, and it tries to inherit from the string interface.

> but since the SafeString object inherits from String.prototype, the length property is read-only

Shouldn't that be fine? You don't ever want to set that property.

I guess I'm a little confused by your original message, Everything should work fine with SafeString. In fact your changes have a bug: that strip out the SafeString interface so now these filters won't *return* a SafeString if one was passed in. See `copySafeness`.

src/filters.js
```diff
@@ -3,6 +3,21 @@
...
 var r = require('./runtime');
 
+
+var normalize = function(value, defaultValue) {
+
+    if (value === null || value === undefined) {
+        return defaultValue;
+    }
+
+    if (value instanceof r.SafeString) {
```
jlongster: Please remove this block. The `normalize` function should only look like this:

```js
function normalize(v, default) {
  if(v === null || v === undefined || v === false) {
    return default;
  }
  return v;
}
```

The filters should work with a SafeString, unless something is amiss.

jlongster:
Hm, actually I think you uncovered a few bugs with `SafeString`. We need to fix those instead of stripping them in the filters though.

First I'm going to copy the `length` property when a `SafeString` is created. I would use getters but I think we want to at least allow IE8 users to use nunjucks with a shim, and you can't shim getters. I think this will make `SafeString` work properly (it can be used transparently as a string)

jlongster:
Oh, I see what you mean about declaring a `length` property: the String prototype catches it and complains that it's not writable. We can fix that by declaring a `length` property for ourselves:

```js
SafeString.prototype = Object.create(String.prototype, {
    length: { writable: true, configurable: true, value: 0 }
});
```

jlongster:
I merged your work but went ahead and fixed `normalize` function myself, thanks!

pierreant-p:
nice thanks!

PR #405, opened 2015-04-09
Update to stable chokidar.

paulmillr:
Lots of improvements & much more stable.


jlongster:
Awesome, thanks! will try it out

PR #417, opened 2015-04-27
Update the test command in the README

tofumatt:
It says `make test`, but there's no `Makefile`. Inspected the `package.json` and saw it should be `npm test`.


jlongster:
Good catch, thanks!

Issue #196, opened 2014-02-20
Slim build should not include WebLoader


mattbasta:
The slim build can't compile templates, so it shouldn't be able to fetch uncompiled templates.

jlongster:
the web loader is 4k *not* gzipped so closing this for now. open a PR if you want it changed

jlongster:
with recent precompiled changed, we can easily do this

Issue #194, opened 2014-02-19
make separate loader for precompiled templates


sunflowerdeath:
If i replace WebLoader with my own loader, precompiled templates are not available.

Usage of precompiled templates does not mean than I want to use WebLoader.
For example, i can have my templates in `<script>` tag on the page, not somewhere in the web, but in the same time I want to use some templates in precompiled way.

If u will make separate loaders for precompiled templates and for web loading, it will be possible to use precompiled templates loader with my own loaders together.

Here's example of separate loaders: https://gist.github.com/sunflowerdeath/9093146

jlongster:
closing, please open a PR if you still have ideas

jlongster:
actually we have found some use cases, now there is a builtin PrecompiledLoader

PR #418, opened 2015-04-28
add wrapper CLI option for loading a precompile plugin

jlongster:
This is my replacement PR for #337. This allows you to specify a wrapper at the precompile CLI like `nunjucks-precompile -w cjs foo.html`, and it will `require` the module `nunjucks-cjs` and use the `wrapper` function is exports as the wrapper option. It can then wrap each template code however it likes.

This PR also adds a new loader that will take a bundle of precompiled templates in node and load them. Any wrapper can emit code to build up an object keyed off of template name and give it to this loader.

### DIFF
--- a/bin/precompile
+++ b/bin/precompile
@@ -36,6 +36,10 @@ var optimist = require('optimist')
         .default('exclude', [])
         .alias('x', 'exclude')
 
+    .describe('wrapper', 'Something something change')
+        .string('wrapper')
+        .alias('w', 'wrapper')
+
     .demand(1);
 
 var argv = optimist.argv;
@@ -51,13 +55,16 @@ lib.each([].concat(argv.filters).join(',').split(','), function(name) {
     env.addFilter(name.trim(), function() {}, true);
 });
 
-console.log(precompile(argv._[0], {
+if(argv.wrapper) {
+    argv.wrapper = require('nunjucks-' + argv.wrapper).wrapper;
+}
 
+console.log(precompile(argv._[0], {
     env : env,
     force : argv.force,
     name : argv.name,
+    wrapper: argv.wrapper,
 
     include : [].concat(argv.include),
     exclude : [].concat(argv.exclude)
-
 }));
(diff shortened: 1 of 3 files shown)


rhengles:
This is great, there's only one suggestion I want to make: Make it possible to use the custom wrapped templates on the Web Loader, maybe setting directly the "this.precompiled" variable or make the PrecompiledLoader available in the browser build.

Suppose we have two applications in the same page which use nunjucks. Their precompiled namespaces should not be shared.

rhengles:
Maybe this could also solve #194 and #196 ?

jlongster:
Yeah, we can definitely use this PR to discuss improvements. Two applications on the same page using nunjucks? That sounds crazy! But I'll think about how to organize this a little better to give more control.

jlongster:
made a new issue for normalizing the precompiled loader: https://github.com/mozilla/nunjucks/issues/426

jlongster:
Alright, I reworked things so there exists a `PrecompiledLoader` both in the node & web worlds. For the web everything should still work as before, as we will by default install one that loads from `nunjucksPrecompiled` if it exists.

The `WebLoader` doesn't know anything about precompiled templates, so I'm going to work on removing it from the slim build.

jlongster:
This is in good shape now; WebLoader is not included in slim builds and you can use the `PrecompiledLoader` from anywhere.

hiaselhans:
How is async-loading precompiled templates handled? Is there a downside of having precompiled templates loaded on-request?
When running dev/deployment environments, this would unify the behaviour a bit (seems to me..) by not having to define(["template",...]) all the time, but have the loader handle this...

jlongster:
What do you mean by "async loading precompiled templates"? Do you mean fetching the actual `precompiled.js` file which contains all of the templates? Nunjucks is not going to be opinionated how you load JavaScript files. You will have to write your code in a way that blocks for the request anyway, it would not get rid of `define`.

hiaselhans:
hi there, thanks for your message..

What i target is to load precompiled templates (not concatenated to one file) at runtime using the same api as with non-rendered... something like env.render("index.html") would load index.js, then cache and render that...
But could be that is the wrong approach anyhow...

PR #423, opened 2015-04-30
export all vars created with `set` even in child templates (fixes #410)

### LINKED ISSUE #410: Import of context issue
Hi

According to #319, import with context should work. But or I'm getting it's purpose in wrong way, or it doesn't work.

I've made branch with demo:

https://github.com/ArmorDarks/grunt-nunjucks-2-html/tree/feature/import/tests/import

Expected output of input2.html is following:

``` html
...

<p>parent's testVar: testVar from page1 baseTestVar from base</p>
<p>parent's testVar2: testVar2 from page1 baseTestVar from base</p>

...
```

while with `{% import "tests/import/input1.html" as parent with context %}` it looks like this:

``` html
...

<p>parent's testVar: </p>
<p>parent's testVar2: </p>

...
```

Quite strange that if import would be specified as `{% import "tests/import/input1.html" as parent %}`, it will output `undefined`:

``` html
...

<p>parent's testVar: testVar from page1 undefined</p>
<p>parent's testVar2: testVar2 from page1 undefined</p>

...
```

jlongster:
@garygreen @carljm does this look right? this will fix https://github.com/mozilla/nunjucks/issues/410

basically we only exported vars created by `set` if the template didn't have a parent scope. This worked in templates that created macros because by default (I think it's default) they don't have access to the scope of the template including it.

However, when using `with context`, there is a parent scope, so variables weren't being exported.

I had to mark certain frames as "top-level" because we don't want to export variables created within a frame that's not the top-level (like inside a `for` loop)


src/compiler.js
```diff
@@ -529,16 +529,20 @@ var Compiler = Object.extend({
...
+            this.emitLine('}');
+
             if(name.charAt(0) !== '_') {
-                this.emitLine('context.addExport("' + name + '");');
+                this.emitLine('if(frame.topLevel) {');
+                this.emitLine('context.addExport("' + name + '", ' + id + ');');
+                this.emitLine('}');
             }
-            this.emitLine('}');
+            //this.emitLine('}');
```
carljm: this should be removed

carljm:
Looks reasonable to me. It'd be good to have a test for the new behavior.

src/environment.js
```diff
@@ -275,7 +275,14 @@ var Environment = Obj.extend({
...
     init: function(ctx, blocks) {
-        this.ctx = ctx;
+        // Make a duplicate of ctx
+        this.ctx = {};
+        for(var k in ctx) {
+            if(ctx.hasOwnProperty(k)) {
+                this.ctx[k] = ctx[k];
+            }
+        }
+
```
jlongster: This change kind of scares me, but it's necessary for proper contexts per-template for imports/includes, etc.

jlongster:
Writing a test did expose a problem that forced me to make that scary change above. Basically a new `Context` is created per template, but it inherits the context where it came from (unless it's importing without context).

Exports are kind of weird: any top-level variables or macros are implicitly exported. But this assumes that at the top-level it's setting these on a context.

Unfortunately that means using `set` in imported templates where context should not have been available will still leak changes, because it's changing the internal variables object which is the same instance as the parent context. We need to *copy* the object instead of just using the same instance.

I don't really think this should break anything, and should make our with/without context environments saner, but it's still a little scary. Anyone care to run this on your project?

carljm:
I wouldn't mind playing around with this a bit more (and comparing the behavior to Jinja2), but I won't have time to do it today. If you want to just merge without waiting for that, that's fine with me.

jlongster:
I'll wait until tomorrow to merge it in, in case you have time tomorrow. I'm going to ask everyone to run master soon anyway because there's lots of changes

jlongster:
Heh, whoops, I forgot that my webpack branch was based on this work so when I pushed that out it pushed this out too. Oh well.

PR #428, opened 2015-05-01
add "loose" option to default filter, add warning about change

jlongster:
@carljm how does this look? I basically warn people about the change and tell them to check their filters. if they are sure it's not going to affect them they can ignore the warning.

I also am not checking specifically for `false` when it is "loose", but any falsy JS value will return the default. Is that sane?

That's the kind if thing that might differ under a "jinja" compat mode... truthiness is the hardest thing to reconcile.

### DIFF
--- a/docs/templating.md
+++ b/docs/templating.md
@@ -702,13 +702,37 @@ If `tags` was `["food", "beer", "dessert"]`, the above example would output `foo
 
 ## Builtin Filters
 
-Nunjucks has ported most of jinja's filters (click through for documentation):
+Nunjucks has ported most of jinja's filters, and has a few of its own.
+We need to work on our own documentation for filters. Some of them are
+documented below, for the rest you can click through to jinja's site.
+
+### default(value, default, [loose])
+
+(aliased as `d`)
+
+If `value` is `null` or `undefined`, return `default`, otherwise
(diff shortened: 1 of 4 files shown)


docs/templating.md
```diff
@@ -702,13 +702,37 @@ If `tags` was `["food", "beer", "dessert"]`, the above example would output `foo
...
-Nunjucks has ported most of jinja's filters (click through for documentation):
+Nunjucks has ported most of jinja's filters, and has a few of its own.
+We need to work on our own documentation for filters. Some of them are
+documented below, for the rest you can click through to jinja's site.
+
+### default(value, default, [loose])
+
+(aliased as `d`)
+
+If `value` is `null` or `undefined`, return `default`, otherwise
```
carljm: why does strict mode also use the default for `null`? this isn't consistent with Jinja's strict default filter, which only uses the default if the variable is undefined.

carljm:
Seems fine to me, apart from the comment about `null`

docs/templating.md
jlongster: `null` and `undefined` usually are treated as the same value. Everywhere we check for `undefined` we also check `null`.

But I see what you mean, this filter is meant to actually only return the default if the variable doesn't even exist in the scope. We can change it to be that.

PR #433, opened 2015-05-04
Added bundle option to precompiled wrappers

rhengles:
Hello!

I was writing wrappers for the major module formats, but with ES6 I had some difficulty. For named exports, the names must be valid variable identifiers. So I couldn't simply export the template name, because it might contain slashes, dashes, etc, so I had to create a plain old JS object to export as default.

But as the file might contain many templates inside, I'd have to create a check for the existence of this object on each template. Instead, I created this PR to be able to create the object once and export it once.

### DIFF
--- a/src/precompile-global.js
+++ b/src/precompile-global.js
@@ -1,19 +1,23 @@
 'use strict';
 
-function precompileGlobal(name, template, opts) {
+function precompileGlobal(templates, opts) {
+    var out = '', name, template;
     opts = opts || {};
 
-    name = JSON.stringify(name);
+    for ( var i = 0; i < templates.length; i++ ) {
+        name = JSON.stringify(templates[i].name);
+        template = templates[i].template;
 
-    var out = '(function() {' +
-        '(window.nunjucksPrecompiled = window.nunjucksPrecompiled || {})' +
-        '[' + name + '] = (function() {\n' + template + '\n})();\n';
+        out += '(function() {' +
+            '(window.nunjucksPrecompiled = window.nunjucksPrecompiled || {})' +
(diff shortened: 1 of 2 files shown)


jlongster:
That's what the standard precompiler does: https://github.com/mozilla/nunjucks/blob/master/src/precompile-global.js

It's nice for the browser because you could potentially load templates in multiple chunks (load a separate file later). I guess that doesn't really apply here though.

rhengles:
Yeah, so the standard precompiler could also benefit from this improvement - reference the object only once in a closure.

jlongster:
I'm not sold that it's a big improvement for the browser, but this looks fine for other environments where there's not a dependable global object like `window`.

jlongster:
I think `bundle` is a little confusing, can you think of a name that indicates some sort of action to run *post*-compiling? `postProcess` ?

rhengles:
Sure, done! :D

rhengles:
I did a little refactoring here (it's best to do this before we ship). The script now precompiles all the templates into an array, then pass this array to the wrapper. The postProcess was a hack, this is a much more proper solution.

jlongster:
Alright, makes sense

rhengles:
For anyone interested, I just published wrappers for the following module formats:

format | github | npm
---|---|---
CJS | [github](https://github.com/rhengles/nunjucks-cjs) | [npm](https://www.npmjs.com/package/nunjucks-cjs)
AMD | [github](https://github.com/rhengles/nunjucks-amd) | [npm](https://www.npmjs.com/package/nunjucks-amd)
ES6 | [github](https://github.com/rhengles/nunjucks-esm) | [npm](https://www.npmjs.com/package/nunjucks-esm)

PR #438, opened 2015-05-11
Added Jinja2 style express string concatenation with tests

campbellwmorgan:
Nunjucks is missing the Jinja2 '~' which expressly concatenates irrespective of source type unlike the javascript '+' which only concatenates if the first operand is a string.


campbellwmorgan:
Apologies - did pull request from my master branch by mistake - closing and re-opening from another branch

PR #448, opened 2015-05-28
Fix/raw

### LINKED ISSUE #447: Raw is broken, gets confused with "fake" variable declarations
Raw blocks break when there's something that looks like a variable declaration but isn't, for example `{{ x }` (it opens like a variable but doesn't end like one).

### Sample

```html
{% raw %}
<code>
{{ i_am_a_dumb_variable }}

{{ i_am_an_imposter_i_kill_nunjucks }
</code>
{% endraw %}
```

### Stacktrace

```
/Users/aaron/git/nunjucks/src/environment.js:425
            else throw err;
                       ^
Template render error: (/Users/aaron/git/nunjucks/tests/raw.html)
  expected endraw, got end of file
    at Object.exports.prettifyError (/Users/aaron/git/nunjucks/src/lib.js:34:15)
    at Obj.extend.render (/Users/aaron/git/nunjucks/src/environment.js:423:27)
    at /Users/aaron/git/nunjucks/src/environment.js:270:35
    at createTemplate (/Users/aaron/git/nunjucks/src/environment.js:194:25)
    at handle (/Users/aaron/git/nunjucks/src/environment.js:206:25)
    at /Users/aaron/git/nunjucks/src/environment.js:223:21
    at next (/Users/aaron/git/nunjucks/src/lib.js:198:13)
    at Object.exports.asyncIter (/Users/aaron/git/nunjucks/src/lib.js:205:5)
    at Obj.extend.getTemplate (/Users/aaron/git/nunjucks/src/environment.js:202:17)
    at Obj.extend.render (/Users/aaron/git/nunjucks/src/environment.js:262:14)
```

This obviously isn't the expected behavior. I'm looking into the parser's raw parsing, but maybe someone more familiar with the codebase knows precisely where it messes up.

AaronO:
This PR fixes #447 

### Summary

* [x] Add unit tests for broken blocks in raw
* [x] Implement a fix in `src/parser.js`

### Bonus
* [x] Add tests for raw blocks in raw blocks
* [x] Support raw blocks in raw blocks

Ideally I think this should be fixed in `src/lexer.js` at the Tokenizer level. (Maybe I'm wrong on that, someone with a better knowledge of the codebase should know).

### DIFF
--- a/src/lexer.js
+++ b/src/lexer.js
@@ -385,6 +385,18 @@ Tokenizer.prototype._extractMatching = function (breakOnMatch, charString) {
     return '';
 };
 
+Tokenizer.prototype._extractRegex = function(regex) {
+    var matches = this.currentStr().match(regex);
+    if(!matches) {
+        return null;
+    }
+
+    // Move forward whatever was matched
+    this.forwardN(matches[0].length);
+
+    return matches;
+};
+
 Tokenizer.prototype.is_finished = function() {
     return this.index >= this.len;
 };
@@ -407,6 +419,12 @@ Tokenizer.prototype.forward = function() {
     }
 };
 
+Tokenizer.prototype.backN = function(n) {
+    for(var i=0; i<n; i++) {
+        this.back();
+    }
+};
+
 Tokenizer.prototype.back = function() {
     this.index--;
 
@@ -426,9 +444,18 @@ Tokenizer.prototype.back = function() {
     }
 };
 
+// current returns current character
 Tokenizer.prototype.current = function() {
     if(!this.is_finished()) {
         return this.str.charAt(this.index);
+    }
+    return '';
+};
+
+// currentStr returns what's left of the unparsed string
+Tokenizer.prototype.currentStr = function() {
+    if(!this.is_finished()) {
+        return this.str.substr(this.index);
     }
     return '';
 };
--- a/src/parser.js
+++ b/src/parser.js
@@ -124,6 +124,8 @@ var Parser = Object.extend({
         else {
             this.fail('expected block end in ' + name + ' statement');
         }
+
+        return tok;
     },
 
     advanceAfterVariableEnd: function() {
--- a/src/parser.js
+++ b/src/parser.js
@@ -549,94 +549,46 @@ var Parser = Object.extend({
     },
 
     parseRaw: function() {
-        this.advanceAfterBlockEnd();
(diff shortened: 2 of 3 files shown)


SamyPesse:
:+1:

adamhavel:
Does this fix a similar problem with broken Nunjucks comments in raw block? For example:

```
{% raw %}{# test {% endraw %}
```

This statement throws an error, too.

AaronO:
@adamhavel The `Tokenizer` is fairly strict, which generally is a good thing, since it makes errors easier to spot and fix.

The problem is that `raw` blocks use the `Tokenizer` so it actually tries to semantically parse the contents of the raw block which triggers all these errors since it's strict and we really don't care about the syntax validity inside raw blocks, the ideal solution would be to not use the tokenizer inside a `raw` block and be lower level.

However it's not that simple because the rest of the parser operates on top of the Tokenizer, so making `raw` blocks an exception isn't easy given the current structure and abstractions in place.

I'll have to dive into the code and figure out how to refactor it in a way that's clean and improves the robustness of `{% raw %}` blocks, but somebody on the `nunjucks` team definitely has a better knowledge of the codebase, so it would be great if thy could jump in.

AaronO:
@adamhavel I added a unit test for your case in my code.

And more importantly I rewrote `Parser.parseRaw` so that it's a lot more robust and in fact far simpler (it's well commented and is still 10 lines shorter than the original while being far more robust).

It passes all the unit tests and we shouldn't have any strange issues with raw blocks again.

AaronO:
@jlongster This should be good to review.

jlongster:
@AaronO thanks a bunch! currently overloaded because of a work week coming up (next week), but I'll try to get through this soon

AaronO:
@jlongster Have you had a chance to take a look at it yet ?

Let me know if you have any questions. Overall this PR simplifies the `raw` block support and makes it more robust, I don't really think there are any regressions (I added extra tests, it passes all those and the old ones of course).

It would be nice to move back on to the `master`: https://github.com/GitbookIO/gitbook/blob/master/package.json#L16 ;)

jlongster:
@AaronO Haven't had the best time maintaining open source projects this year, sorry. This looks great though! Merging.

AaronO:
@jlongster No problem, it happens to all of us at times :) Thanks for merging the PR ! :+1:

PR #457, opened 2015-06-10
Add a new browserCache option for the web loader

borgstrom:
My usage of nunjucks results in a scenario where I cannot precompile templates and I would prefer not to have the browser load a new copy of the template each time it is rendered, so this adds a new option that will disable the addition of "s=<timestamp>" to the URL.


jlongster:
How is this different then passing `useCache: true` to the browser loader, which will store the templates in memory and cache them by name?

Also, I don't want to force the user to pass in the string to append. They shouldn't have to think about that. It should be on or off.

borgstrom:
It's different because I'm trusting the server and the browser to tell me when a template changed through standard http caching mechanisms rather than blindly caching everything.

I'm not forcing anyone to specify anything:

`browserCache: false` (the default) will append `s=<date>`, like it works today.
`browserCache: true` will omit the s= param altogether.
`browserCache: "my string"` will append `s=my+string`

The last thing is that `useCache` is not well documented:

```
#1003 ❯❯❯ curl http://mozilla.github.io/nunjucks/api.html | grep useCache
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 65700  100 65700    0     0  3195k      0 --:--:-- --:--:-- --:--:-- 4010k

#1004 ❯❯❯ 
```

I did find it here https://github.com/mozilla/nunjucks/blob/master/docs/api.md, but it's strange it's not in the public docs.

borgstrom:
And, for reference here's the use case of how I'm using this new option.

    web: { browserCache: DEBUG ? false : VERSION }

jlongster:
It's because those options were updated in the latest version, which just got pushed out. It's not on the public docs because 2.0 was just published: http://mozilla.github.io/nunjucks/api.html#configure

jlongster:
I'm open to this, just thinking it through. I see what you mean. Are you serving your templates with specific caching headers? Usually browsers cache the templates quite aggressively, which is why we add the random query.

carljm:
I think this is a reasonable feature and am willing to merge it. The patch needs tests and documentation.

vecmezoni:
@borgstrom could you please provide tests and docs for your PR?

PR #465, opened 2015-06-22
test definition of global macros from within a simple include file

boutell:
This test currently fails with this error:

Error: Unable to call `foo`, which is undefined or falsey

### DIFF
--- a/tests/compiler.js
+++ b/tests/compiler.js
@@ -466,6 +466,14 @@
             finish(done);
         });
 
+        it('should allow definition of global macros in an included file', function(done) {
+          equal('{% include "include-macro.html" %}' +
+            '{{ foo("test") }}',
+            'testtest'
+          );
+          finish(done);
+        });
+
         it('should compile call blocks', function(done) {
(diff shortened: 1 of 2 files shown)


garygreen:
This doesn't work in jinja and I don't think we will support it. If you want to use macros you should `import` them.

boutell:
... Oops. Apparently we've been doing it wrong from day one, then. But the
jinja docs support what you're saying, yes. We'll just have to go modern
before we can go 2.0.

On Sun, Jul 5, 2015 at 10:04 AM, Gary Green <notifications@github.com>
wrote:

> Closed #465 <https://github.com/mozilla/nunjucks/pull/465>.
>
> —
> <https://github.com/mozilla/nunjucks/pull/465#event-348033694>.
>



-- 


*THOMAS BOUTELL, *DEV & OPS
P'UNK AVENUE | (215) 755-1330  |  punkave.com

ArmorDarks:
I'm a bit confused. So, does Jinja support it, or not?

Issue #344, opened 2014-12-23
Strict/non-strict syntax modes


bgmort:
There seem to be two big reasons people use nunjucks over competing template engines:  for Python/Jinja compatibility, or because it's the best javascript-powered template engine around.  While jinja compatibility seems to be a priority, a lot of useful improvements could be made for nunjucks as a javascript library that would not be jinja compatible, but help it be the best possible template engine for javascript.  That's why I want to throw out the idea of having a 'strict' mode that is on by default, but when turned off, would allow language features that are not a part of jinja.  For example, in non-strict mode, the familiar `&&`, `||`, and `!` operators could be used instead of `and`, `or`, and `not`.  The ternary operator could also be used instead of Python's odd `.. if .. else ..` syntax.  I think this could go a long way toward making nunjucks the template engine of choice for the growing numbers of all-javascript projects.

garygreen:
I'm not sure this would really add much value. We do aim to have as close API/syntax to jinja as much as possible, so adding `&&`  `||` `!`, although familiar to javascript, would potentially polute and confuse, having multiple ways of doing the same thing. -1 for me

ArmorDarks:
I must agree with @garygreen 

:-1: for me too

If someone wants Nunjucks features with JS operators — it's always possible to make a fork.

eddywashere:
As someone new to nunjucks and switching from swig, this would have been incredibly valuable. Instead I spent a lot of time figuring out what wasn't supported. Please consider revisiting this option, as it would be great for first users and people coming from swig. I would -1 to forking because that does not help the community. It only dissolves it.

PR #474, opened 2015-07-02
Add new filter: contains

GochoMugo:
```
contains(container, item)
```
Return `true` if the container contains the item. Otherwise `false`.

If the container is an Array, the function loops through the container comparing each element with the item using `===`, until the item is found or all elements are compared, whichever comes first.

If the container is an Object, the function checks if `container[item]` is **not** `undefined`.

If the container is a String, the function uses `String.prototype.indexOf` to find the item.

Finally, if the container is of any other type, the function throws a `TemplateError`.

The function has an alias, namely, `has`.

Issue: forfutureLLC/nunjucks#2

Any suggestions?

### DIFF
--- a/src/filters.js
+++ b/src/filters.js
@@ -534,11 +534,33 @@ var filters = {
     'int': function(val, def) {
         var res = parseInt(val, 10);
         return isNaN(res) ? def : res;
+    },
+
+    contains: function(container, item) {
+        if (container instanceof Array) {
+            for (var index = 0; index < container.length; index++) {
+                if (container[index] === item) {
+                    return true;
+                }
+            }
+            return false;
+        }
+
+        if (typeof container === 'object') {
+            return container[item] !== undefined;
+        }
+
+        if (typeof container === 'string') {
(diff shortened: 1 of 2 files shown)


ricordisamoa:
It is my understanding that Nunjucks tries to be as close to Jinja2 as possible, and I don't think the latter has a `contains` filter.

How would `container|contains(item)` differ from `item in container`?

GochoMugo:
Ooh! I didn't realize we would have to learn using Jinja2 to use Nunjucks. Found the `in` operator [here](http://jinja.pocoo.org/docs/dev/templates/#other-operators). This PR just got useless.

Just a Question, does that mean no more features shall be added unless Jinja2 does?

ricordisamoa:
https://github.com/mozilla/nunjucks/issues/344#issuecomment-70645452
> We do aim to have as close API/syntax to jinja as much as possible

https://github.com/mozilla/nunjucks/pull/304#issuecomment-59860413
> we don't guarantee 100% jinja compatibility, and we diverge where we think is helpful

PR #476, opened 2015-07-05
Add getGlobal() function

atorkhov:


garygreen:
Do you have a use case for this?

atorkhov:
Yes, I'm doing unit tests and checking custom filters and global functions behaviour. To get filters I can use getFilter(), but for globals I need to expose them in my application somehow.

garygreen:
Ok cool, would you mind adding some tests in `tests/global.js` for this?

atorkhov:
Updated pull request with test and docs.
One question through - are globals really supposed to be same across all enviroments?

devoidfury:
In a node environment, you can get access to the globals with `var nunjucks_globals = require('nunjucks/src/globals.js');`

atorkhov:
I'm using it in browser environment.

atorkhov:
Thanks. Any comment on are globals really supposed to be same across all environments or not?

garygreen:
At the moment globals are the same across all environments.

Issue #478, opened 2015-07-14
Argument from caller macro is incorrectly passed through to callee macro?


jgerigmeyer:
If a macro is called (using ``{% call %}``) with an argument, and the ``caller()`` is itself a macro that accepts an argument with the same name with a default of ``none``, the argument from the ``caller`` macro is incorrectly passed through to the ``callee`` macro. That sounds ridiculously confusing, so here's the test case:

```
{% macro inner(class=none) %}
  <div class="{{ class }}"></div>
{% endmacro %}

{% macro outer(class=none) %}
  <div class="{{ class }}">
    {{ caller() }}
  </div>
{% endmacro %}

{% call outer(class='outer') %}
  {{ inner() }}
{% endcall %}
```

When rendered via jinja2, this is the output:

```
<div class="outer">
  <div class="None"></div>
</div>
```

However, when rendered via nunjucks, the output is:

```
<div class="outer">
  <div class="outer"></div>
</div>
```

This seems to only happen when the default for the ``inner`` arg is ``none``; it doesn't occur when the default is ``''`` or ``false``.

carljm:
Interesting! I'll look into this. I wonder if it's just arguments, or works for any variable of a matching name in the calling namespace...

carljm:
This was partly fixed by #667 -- macro frames are now totally separated from the calling frame, so there's no longer any way for a macro to reach out and see or touch its calling frame. So you can no longer see this behavior with macro arguments with a default of `none`. Looking into whether it can still crop up in other situations...

jgerigmeyer:
I'm still seeing this bug (or seeing it again?):

http://jsfiddle.net/jgerigmeyer/tro6jhy9/1/
http://jsfiddle.net/jgerigmeyer/6cL8wqd7/1/

PR #480, opened 2015-07-21
Add 'none' as a literal evaluating to null.

carljm:
I ran across this in working on the fix for #478.

Jinja2 has three special constants: `true`, `false`, and `none` (mapping to Python's `True`, `False`, and `None`, respectively).

Nunjucks has `true` and `false` (mapping to JS `true` and `false`), but no explicit support for `none`.

This has gone unaddressed thus far, because in practice a Jinja template using `none` ports just fine to nunjucks anyway, as the symbol `none` evaluates to `undefined`, which behaves similarly-enough to Python's `None` (mostly just in that it's falsy) to make no real difference.

But there are some subtle problems with this.

First, it's pointlessly inefficient. There's no reason for use of `none` to require a `contextOrFrameLookup` at runtime every place it is used.

Second, it allows for the name `none` to be re-bound to some other value, resulting in confusing code and needless Jinja incompatibility.

Third, the logical mapping for `none` to JS is really `null`, not `undefined`. Most of the time this makes no difference, but it does make a difference with the `default` filter. In Jinja, `{{ none|default('d', false) }}` evaluates to `None` (because `None` is not an undefined value), whereas in nunjucks it prints `d` (because it is undefined).

This pull request turns `none` into a lexed constant just like `true` and `false`, with the value `null`.

This could be backwards incompatible, if someone is relying on the current behavior of `none` with the `default` filter, or is re-binding the name `none` to some other value. I think this backwards-incompatibility is acceptable as a bugfix, and the time to do it is now (before 2.0 is released).


SamyPesse:
@carljm This PR looks ready to be merged

carljm:
@SamyPesse Thanks for the nudge, I think you're right. Merging now.

PR #482, opened 2015-07-23
Update web-loaders.js

aksian:
As declared in API docs (http://mozilla.github.io/nunjucks/api.html#webloader) baseURL is the URL to load templates from (must be the same domain), and it defaults to the current relative directory. But up to now it defaults to domain root. This is an update to ensure that it worked as expected.


jlongster:
Thanks!

Issue #471, opened 2015-06-29
"replace" filter (and other string-based default filters) not working with undefined strings


ohmance:
I'm using nunjucks as an email templating library where I can put in variables that correspond to user data.  Say a user looks something like this:

{ "first_name": "Bob", "last_name":"Dylan" }

Sometimes a user has a middle name, sometimes they don't have a first name, etc.

I'm attempting to use the "replace" filter (and various other string-based filters).  It works well in most cases unless the property doesn't exist for the user, then it throws an error because the input property is undefined.

Look at nunjucks.js line 4249 where the replace method is defined.  The parameter str will be undefined and thus not have the "replace" method in it's prototype if the input parameter is undefined.

I suggest that a better experience for this and other string-based filters is to add a line line this as the first line:
`str = (str || '')`

This would silently create an empty string if one is not defined.

jlongster:
Can you try on master? I believe the filters have been updated. For example `replace` has all these tests for bad inputs: https://github.com/mozilla/nunjucks/blob/master/tests/filters.js#L307

I'm going to push out 2.0 soon which includes these fixes.

carljm:
Closing due to lack of confirmation that the issue still exists.

PR #494, opened 2015-08-19
Copysafeness handling undefined & null gracefully

RickCraig:
This is a simple fix that will not throw and error when an ```undefined``` or ```null``` is passed into the ```runtime.copySafeness()``` method.

Instead of an error, an empty string will be passed back. To use issue #471 as an example:
```javascript
var template = 'Hello {{firstName | string}} {{middleName | replace(',', ' ')}}{{lastName | string}}';
nunjucks.renderString(template, {
  firstName: 'Sammy',
  lastName: 'Jones'
});
// before - Error Thrown
// after - 'Hello Sammy Jones'
```

### DIFF
--- a/src/runtime.js
+++ b/src/runtime.js
@@ -176,7 +176,9 @@ function copySafeness(dest, target) {
     if(dest instanceof SafeString) {
         return new SafeString(target);
     }
-    return target.toString();
+    var type = typeof target;
+    return type !== 'undefined' && type !== 'null' ?
+        target.toString() : '';
 }
 
 function markSafe(val) {


jlongster:
I don't think that will throw anymore on master. All of the filters have been updated to gracefully handle bad inputs. There are even tests for replace: https://github.com/mozilla/nunjucks/blob/master/tests/filters.js#L307

`copySafeness` should do just one thing. We want to be very careful where handle `null` and `undefined` errors, and make filters explicitly handle that if they want to. Because otherwise templates will swallow errors too much, and users will be wondering why nothing is showing when something should, when a filter should have been throwing an error.

RickCraig:
Ok, that makes sense, closing. Thanks.

PR #502, opened 2015-08-30
Add failing test for multiple {% raw %} blocks

ricordisamoa:

PR #503, opened 2015-09-01
Fix/multiple raw

AaronO:
Improves new raw block logic (more robust)

### Fixes:
* Greedy raw matching that broke multiple raw blocks

Fixes #502


carljm:
LGTM, merging. Thanks!

AaronO:
Thanks :+1:

PR #504, opened 2015-09-01
Catch correctly error from loaders

SamyPesse:
When an async loader return an error, Nunjucks is crashing the node process because of this [line](https://github.com/mozilla/nunjucks/blob/master/src/environment.js#L225).

This PR fixes this, and add a test for it.


carljm:
LGTM. Merging.

Issue #255, opened 2014-07-03
Supporting keyword arguments for sort filter


patrickarlt:
Yesterday I ran into an issue trying to use some of Nunjucks builtin filters

According to the Jinja docs the [`sort`](http://jinja.pocoo.org/docs/templates/#sort) filter is able to accept keyword arguments for sorting arrays of objects

```html
{% for post in posts|sort(attribute="date") %}
  <h3>{{ post.title }} <small>{{post.date}}</small></h3>
{% endfor %}
```

Which simply leaves the array of blog posts unsorted. Taking a quick look at the [tests for sort](https://github.com/mozilla/nunjucks/blob/master/tests/filters.js#L264) it looks like filters don't support keyword arguments but this does work ...

```html
{% for post in blog|sort(false, false, "title") %}
  <h3>{{ post.title }} <small>{{post.date}}</small></h3>
{% endfor %}
```

So my question is should the built in filters be documented in the Nunjucks site (rather then just linking to Jinja) I could move all the docs easily. Or should filters be modified to accept keyword arguments?

ArmorDarks:
probably code should be updated to correspond latest Jinja changes regarding `sort` keywords

jlongster:
We've been meaning to document our filters and not rely on jinja's documentation. If someone wants to start that work, that would be great. And yes, we could probably improve `sort` by using keyword arguments.

PR #510, opened 2015-09-07
kwargs support for sort filter

pardo:
Added support for kwargs on the sort function 

http://jinja.pocoo.org/docs/dev/templates/#sort

https://github.com/mozilla/nunjucks/issues/255


SamyPesse:
Thanks :+1:

PR #533, opened 2015-09-28
Fix infinite loop after an undefined function called

tdelmas:
In some complex case, an error throw because of the call of an undefined function 
```
{{ does_not_exist() }}
```
Is not propagated, but instead end up in an infinite loop.

(it's my first contribution here, feel free to ask some cleanup if necessary before accepting the pull request)


tests/compiler.js
```diff
@@ -1125,5 +1125,35 @@
                   'may the forth be with you');
             finish(done);
         });
+        
+
+        it('should thrown an error', function(done) {
```
carljm: Can we get a more specific description of the previously-broken case here? Like `it('should throw an error when including a file that calls an undefined macro')`?

tests/compiler.js
```diff
@@ -1125,5 +1125,35 @@
...
+                function(err, res) {
+                    expect(res).to.be(undefined);
+                    expect(err).to.match(/Unable to call `\w+`, which is undefined or falsey/);
+                }
+            );
+
+            finish(done);
+        });
+
+        it('should thrown an error', function(done) {
```
carljm: And again here. Definitely don't want two cases with exactly the same description; makes the test-runner output much less useful if one of them fails.

carljm:
Looks great, thanks for the contribution! If you're willing to make a small change to the test descriptions as noted above, and then squash down to one commit (rather than two commits with identical commit message, where it looks like the second just adds some forgotten files), I'd be happy to merge this.

(If you don't feel comfortable enough with git to handle squashing down to one commit, that's fine, just let me know and I can squash while merging.)

tdelmas:
Sure, I'll fix that tomorrow!

PR #534, opened 2015-09-29
Fix infinit loop when calling an undefined function

tdelmas:
In some complex case, an error throw because of the call of an undefined function 
```
{{ does_not_exist() }}
```
Is not propagated, but instead end up in an infinite loop.


carljm:
Thanks! Committed in b638d15630dd345b0b12fa78dcbc26a0b4f1caad

tdelmas:
Awesome!

Issue #485, opened 2015-07-29
Env add functions should ideally return env object


robgraeber:
Sometimes when configuring Nunjucks I want to just chain the addFilter / addGlobal functions etc. Does anyone else think it would make sense if addFilter / addGlobal would return the env object? 

So code like this would be possible:
```
Nunjucks.configure(templateDir, {
    autoescape: true,
    trimBlocks: true,
    lstripBlocks: true,
    express: app
}).addGlobal('func1', function(arg) {
    return 'asdf'
}).addGlobal('func2', function(arg) {
    return 'asdf'
}).addGlobal('func3', function(arg) {
    return 'asdf'
}).addGlobal('func4', function(arg) {
    return 'asdf'
});
```

I could do the PR if others thought it was a good idea. These are the functions I was thinking of making chainable:
```
addFilter
addGlobal
addBlock
addExtension
express
```

devoidfury:
Yes, this would be a nice touch, I agree.

jlongster:
Works for me, feel free to open a PR!

PR #537, opened 2015-10-01
Add chaining of env 'add' methods

robgraeber:
As discussed in https://github.com/mozilla/nunjucks/issues/485, adds chaining for env 'add' methods like `addGlobal, addFilter, addBlock, etc`. I also wrote a test for addGlobal chaining and verified all the tests still pass. Let me know if you want anything improved.

@devoidfury @jlongster @carljm 

This would make chaining like this possible:
```
Nunjucks.configure(templateDir, {
    autoescape: true,
    trimBlocks: true,
    lstripBlocks: true,
    express: app
}).addGlobal('func1', function(arg) {
    return 'asdf'
}).addGlobal('func2', function(arg) {
    return 'asdf'
}).addGlobal('func3', function(arg) {
    return 'asdf'
}).addGlobal('func4', function(arg) {
    return 'asdf'
});
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ Changelog
 master (unreleased)
 -------------------
 
+* Add support for chaining of addGlobal, addFilter, etc. Thanks Rob Graeber. Merge of
+  [#537](https://github.com/mozilla/nunjucks/pull/537)
 * Fix error propagation. Thanks Tom Delmas. Merge of
   [#534](https://github.com/mozilla/nunjucks/pull/534).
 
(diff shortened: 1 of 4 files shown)


carljm:
Thanks @robgraeber ! Mind adding an entry to the changelog for this? Also, do you think this is worth mentioning in the documentation for those methods?

robgraeber:
Hey @carljm sure, updated the changelog and made references in the docs. I think mentioning those functions have a return value now could be nice.

carljm:
Merged, thanks for the contribution!

PR #538, opened 2015-10-02
Document use of 'else' in 'for' tags

jgj:
Took some digging to find that this was supported. Adding it to the documentation as it's a handy feature to know about.


carljm:
Thank you! Merged.

Issue #267, opened 2014-08-06
Make empty containers falsy in templates


carljm:
In Jinja2, `{% if [] %}yes{% endif %}` has no output.

In nunjucks, it outputs `yes`.

Is this an area where nunjucks desires to maintain the JS semantics, or would a PR be accepted to make empty containers (anything where the `|length` filter would return `0`) consistently evaluate as false?

ArmorDarks:
Maybe it would be easier to decide if someone will post their use cases when it can be useful and when not.

To be honest, surprisingly, I never bashed it that issue yet.

I must say that `{% if [] %}yes{% endif %}` = `yes` sound right for me, since there is an array, even though it's empty, it exists. And when you're making that kind of check you're in fact just checking existence of variable and does it evaluate to true or false.

However, what I can see that in current implementation of nunjucks it's impossible to check without custom filter whether array is empty or not. If we'll go with Jinja way, it will be at least possible to determinate whether array empty or not without any filters, with something like this (if I'm getting it right):

``` twig
// will output yes
{% set array = [] %}
{% if array or array == [] %}
  yes
{% endif %}`

// won't output yes
{% set array = [] %}
{% if array %}
  yes
{% endif %}`
```

carljm:
You can already check using the built-in `length` filter.

Whether it seems "right" or not probably depends on language background. In Python empty containers evaluate to `false`. In JS they evaluate to `true`. So this is really entirely a choice between Jinja compatibility and JS semantics, which is kind of the permanent existential angst of nunjucks :P

carljm:
(I have considered maintaining an actual nunjucks fork whose explicit goal is as close to full Jinja compatibility as possible.)

howardroark:
Do it :)

jlongster:
@carljm I'd love to see a list of changes that you would make. I think it would be really cool to have a `jinjaCompat: true` option that tweaks the semantics and adds a few things to make it as close as possible. (the reason why I never stated full compat as a goal was because it's a rabbit-hole of emulating Python, which is not only lots of work but really slow)

I'd prefer to merge your changes in here rather than fork, of course. But it's up to you. Also, if you were interested in taking over nunjucks and moving it that direction, I'd be up for talking about that. I want to get the repo in good shape soon and delegate more to contributors.

carljm:
[...]

PR #541, opened 2015-10-02
Bug : you can't test a variable true-ness after an assignement from a macro

tdelmas:
Sorry to not include a fix.

the generated code for the **if** of

```
{% set a = macro_that_return_empty_string(...) %}
{% if ( a ) %}
```

is

```
if( runtime.contextOrFrameLookup(context, frame, "a") ) {
```

which return a SafeString **{ 'val':'','length':0 }** instead of **''** (so the **if** gets always **true**)

It works with **{% if ( a != '') %}** because of the toString method on SafeString.

I think the compiler should produce something like : 

```
if( runtime.contextOrFrameLookup(context, frame, "a").val ) {
```

### DIFF
--- a/tests/compiler.js
+++ b/tests/compiler.js
@@ -1154,5 +1154,15 @@
             finish(done);
         });
 
+        it('\'if\' should behaves exactly as javascript\'s if behaves (if after set)', function(done) {
+            equal('{% set a = \'\' %}{% if a %}true{% else %}false{% endif %}', 'false');
+            finish(done);
+        });
+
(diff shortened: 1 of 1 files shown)


carljm:
This falls into the same general category as #267, in that it would require a new function to be used everywhere we evaluate truthiness or falsiness in order to provide special handling for some objects, rather than just relying on the native JS evaluation of truthiness or falsiness.

I think in the long run for more consistent boolean semantics this is probably what we'll want to do, but there are some concerns about the potential performance impact.

PR #543, opened 2015-10-03
Update getting-started.md

Zabari:
Several small grammatical errors.
"Usage" sounds more professional than "using" and also makes more sense gramatically.


carljm:
Committed with a fix in 30691f8a71323e78978616c48558e44147d307e6 -- thanks!

PR #546, opened 2015-10-06
Fixed sudden "Cannot read property '...'  of undefined"

umaxfun:
Without this line I always have an exception:
TypeError: Cannot read property '<name of html template>' of undefined
    at Obj.extend.getTemplate (/Users/.../node_modules/nunjucks/src/environment.js:179:41)
    at Obj.extend.render (/Users/.../node_modules/nunjucks/src/environment.js:289:14)
    at Object.module.exports.render (/Users/.../node_modules/nunjucks/index.js:71:14)
...

### DIFF
--- a/src/environment.js
+++ b/src/environment.js
@@ -175,6 +175,7 @@ var Environment = Obj.extend({
 
         for (var i = 0; i < this.loaders.length; i++) {
             var _name = this.resolveTemplate(this.loaders[i], parentName, name);
+            if(!this.loaders[i].cache) this.loaders[i].cache = {};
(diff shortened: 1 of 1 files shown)


carljm:
This looks harmless, but I'm confused as to why I (and apparently most other nunjucks users) don't seem to have encountered this problem. Could you show the code you are using to render, so I can try to reproduce? Or alternately, can you add a test to the test suite which fails without this fix and passes with the fix?

carljm:
Closing for now due to lack of feedback.

PR #548, opened 2015-10-07
Make trimBlocks also trim windows style line endings

magnusjt:
When trimBlocks is enabled, only unix-style (LF) line endings are removed from the end of blocks. This pull request makes it so that also windows style (CRLF) line endings are removed.


carljm:
Looks good to me! Thanks for the contribution. Would you mind adding a note to the Changelog file? Then I'll merge.

magnusjt:
Added a note as requested :)

carljm:
Merged, thanks!

Issue #507, opened 2015-09-05
require('nunjucks') within webpack returns an empty object


at0g:
When requiring nunjucks within a webpack project, an empty object is returned instead of nunjucks.

eg.

```
// webpack.config.js
module.exports = {
    entry: './main.js'
};
```

```
// main.js
var nunjucks = require('nunjucks');
console.log(nunjucks); // outputs {}
```

at0g:
The config file below seems to fix the issue - it borrows the plugins configuration from *nunjucks/bin/bundle.js* to create a non-slim web build. It's a lot of configuration to include one package.

```
// webpack.config.js
var webpack = require('webpack');
module.exports = {

    entry: './main.js',

    plugins: [
        new webpack.NormalModuleReplacementPlugin(/nunjucks$/,
            'nunjucks/index.js'),
        new webpack.NormalModuleReplacementPlugin(/(path|fs)$/,
            'node-libs-browser/mock/empty'),

        new webpack.NormalModuleReplacementPlugin(/nunjucks.+precompile$/,
            'node-libs-browser/mock/empty'),

        new webpack.DefinePlugin({
            'process.env': {
                IS_BROWSER: true
            }
        }),
        new webpack.NormalModuleReplacementPlugin(/nunjucks.+loaders\.js$/,
            './web-loaders.js')
    ]
};

```

jgerigmeyer:
Like many bundlers, Webpack uses the ``browser`` field from ``package.json`` for the entry point, so in this case it's pulling in https://github.com/mozilla/nunjucks/blob/master/browser/nunjucks.js. That file doesn't export any modules, but simply defines the local ``nunjucks`` variable. Thankfully, Webpack has a loader specifically for this case: https://github.com/webpack/exports-loader

In your ``webpack.config.js`` file:

```js
module.exports = {
  ...
  module: {
    loaders: [
      {
        test: /nunjucks\/browser\/nunjucks\.js$/,
        loader: 'exports?nunjucks'
      }
    ]
  }
}
```

This will expose the ``nunjucks`` local variable from the ``browser/nunjucks.js`` file whenever it's required.

We could modify nunjucks to add a UMD wrapper or export a variable, but the current approach seems the simplest and most flexible with various bundlers and browser environments.

carljm:
@jgerigmeyer Do you think that's worth adding to the docs somewhere?

at0g:
[...]

PR #556, opened 2015-10-20
change the output library target to commonjs2.

mohannadsh:
In the current version, when I use the nunjucks in the browser using nunjucks = require('nunjucks');, I recieve an empty object, and I cannot use nunjucks to render templates.

The suggested solution is to change the target to commonjs2, so that the nunjucks methods are exported to the module.


carljm:
See discussion on #507.

mohannadsh:
I saw the discussion, I tried the exports-loader and it didn't export the whole nonjucks object, I think that this is a simpler fix, What do you think, do you see any issue with it?

ingro:
@mohannadsh I solved with exports-loader by doing this:
```
{
    test: /nunjucks\.js/,
    loader: 'exports?nunjucks'
}
```

Didn't work with `test: /nunjucks\/browser\/nunjucks\.js$/,`

Rowno:
FYI, exports-loader doesn't help Browserify users.

mohannadsh:
@Rowno 
You're right, exports loader didn't work with me.

mohannadsh:
@carljm 
Are you going to push this change?

carljm:
@mohannadsh I'm not sure. I'm not an expert on JS modules stuff; I'm hoping @jgerigmeyer or another contributor will weigh in on this and #507 and make a decision on where to go with it.

carljm:
(BTW this needs rebasing and a conflict resolved before it can be merged, but you may not want to bother with that until we make a decision about how to handle nunjucks packaging going forward.)

jgerigmeyer:
I'd prefer a UMD wrapper so that it still works natively in the browser. See https://github.com/mozilla/nunjucks/pull/606.

carljm:
@mohannadsh Does that resolution work for you?

Issue #437, opened 2015-05-09
allow suppressing missing-file error on include


jmls:
if you are {% include "someFile.js" %} and someFile.js doesn't exist is it possible to suppress the error and just return an empty string ?

I have a template that makes some includes, and if the included file doesn't exist I don't want the renderer to crap out on me ;)

jlongster:
I'd rather not. Suppressing errors typically is way more frustrating than magically trying to make stuff work. I'd suggest just making sure that file exists but is empty. Open a PR if you want to add an option to the `include` tag to suppress the error, I suppose that could work.

MMore:
I am interested in the same feature, but I am not really familiar with JS and templates for JS at all. I would add an optional parameter like `skipNonExistent`:

`{% include "someFile.js" skipNonExistent %}`

What do you think, @jlongster?

carljm:
Jinja2 has this feature: http://jinja.pocoo.org/docs/dev/templates/#include

I wouldn't have any problem with a pull request implementing it, using the same syntax as Jinja2 (`ignore missing`).

MMore:
It seems I am not deep enough. I am able to do the parsing and I can pass the parameter. But to take care of this parameter and not throw a exception is kind of tricky. Can anyone help me with this part?

carljm:
@MMore I don't have time to look at it in-depth right now, but based on a brief glance it looks to me like `Environment#getTemplate` (in `environment.js`) will need to grow a new boolean parameter to support this feature. The parser should store a new boolean attribute on the include node indicating that missing templates should be ignored, and then in `compiler.js` in `compileInclude` you'd emit the appropriate additional parameter to `getTemplate` depending on the value of the node attribute.

You'll need to check all the loaders (web, filesystem, precompiled) to make sure they return `null` and don't throw an error internally when a template is missing. Then `getTemplate` just needs to return some kind of null template object (whose render method just always returns the empty string) if it gets a null value back from the loader's `getSource` method, and the "ignore missing" flag is set.

That's the direction I'd start in, anyway. Good luck!

carljm:
Fixed with merge of #559

PR #559, opened 2015-10-21
Added feature to ignore missing template files for include

MMore:
This feature belongs to #437.

I will add documentation etc. when anyone of you confirm this as a possible solution. Thank you.


src/parser.js
```diff
@@ -427,7 +427,21 @@ var Parser = Object.extend({
...
+        var tagName = 'include';
+        var tag = this.peekToken();
+        if(!this.skipSymbol(tagName)) {
+            this.fail('parseInclude: expected '+ tagName);
+        }
+
+        var node = new nodes.Include(tag.lineno, tag.colno);
+        node.template = this.parseExpression();
+
+        if(this.skipSymbol('ignoreMissing')) {
```
carljm: the syntax should be `ignore missing` (two words) to match Jinja, not `ignoreMissing`.

src/environment.js
```diff
@@ -193,10 +193,16 @@ var Environment = Obj.extend({
         } else {
             var syncResult;
             var _this = this;
+            var fileMissing = false;
```
carljm: I don't think this new var is needed -- it's effectively equivalent to `!info`.

src/environment.js
```diff
@@ -193,10 +193,16 @@ var Environment = Obj.extend({
...
 
             var createTemplate = function(err, info) {
                 if(!info && !err) {
-                    err = new Error('template not found: ' + name);
+                    if(!ignoreMissing) {
+                        err = new Error('template not found: ' + name);
+                    }
+                    else {
```
carljm: eliminate this `else` clause, not needed

src/environment.js
```diff
@@ -208,11 +214,18 @@ var Environment = Obj.extend({
                     }
                 }
                 else {
-                    var tmpl = new Template(info.src, _this,
+                    var tmpl;
+                    if(ignoreMissing && fileMissing) {
```
carljm: this can be replaced with `if (!info)` -- though probably clearer to swap with the `else` clause, so you have `if (info) { create a real template } else { create a no-op template }`

carljm: there's no need to check `ignoreMissing` again here -- if the template is missing and `ignoreMissing` is not set, there'd have already been an error.

src/parser.js
```diff
@@ -427,7 +427,21 @@ var Parser = Object.extend({
     },
 
     parseInclude: function() {
-        return this.parseTemplateRef('include', nodes.Include);
```
carljm: After this PR, `parseTemplateRef` is only used by `parseExtends`. I think we may as well just remove `parseTemplateRef` entirely in this PR.

carljm:
Looks good! Thanks for diving in here. I think this is pretty close, just made a few small comments. The tests look good; I don't mind not having a parser-specific test added, if it's covered in the compiler test anyway. A couple more things:

1. It looks to me like we don't have tests anywhere for the "missing template" error condition. It would be great to add a test for that case along with the tests for suppressing the error.

2. Can you add a note in the `CHANGELOG.md` file, copying the format of the existing entries?

And of course a documentation note is needed as well.

Thanks again!

MMore:
Thanks for you help and feedback. I will fix it till the end of the week.

tests/compiler.js
```diff
@@ -694,6 +694,38 @@
...
+                function(err, res) {
+                    expect(res).to.be(undefined);
+                    expect(err).to.match(/template not found: missing.html/);
+                }
+            );
+
+            finish(done);
+        });
+
+        it('should include templates that does not exist if the error is suppressed', function(done) {
```
carljm: This seems to duplicate the tests added above. Shouldn't either this `it` block or the two added tests above under `it('should include templates')` be removed?

carljm: The wording here also doesn't seem quite right -- it's not possible to "include templates that does not exist." I would say instead something like `it('should fail silently on missing templates if requested')`

carljm:
Looks great! Thanks for your work on this. Merging.

Issue #39, opened 2012-12-06
Conditional extends should work, no?


pennyfx:
I'm trying to return html fragments and figured that I could achieve this by conditionally setting a template to extend from.   The true case works, but the false case throws an exception.

```javascript
   it('should allow extends blocks inside of if blocks', function(){
        var s = render( '{% if true %}'+
                        '{% extends "base.html" %}'+
                        '{% endif %}'+
                        '{% block block1 %}fragment{% endblock %}');
        s.should.equal('FoofragmentBazFizzle');

        s = render( '{% if false %}'+
                    '{% extends "base.html" %}'+
                    '{% endif %}'+
                    '{% block block1 %}fragment{% endblock %}');
        s.should.equal('fragment');
    });
```

Error

```
TypeError: Cannot call method 'rootRenderFunc' of undefined
      at [object Object].root [as rootRenderFunc] (eval at <anonymous> (/Users/aschaar/programming/xtags-org/node_modules/nunjucks/src/environment.js:273:24))
      at [object Object].render (/Users/aschaar/programming/xtags-org/node_modules/nunjucks/src/environment.js:242:21)
      at render (/Users/aschaar/programming/xtags-org/node_modules/nunjucks/tests/util.js:9:14)
      at Context.<anonymous> (/Users/aschaar/programming/xtags-org/node_modules/nunjucks/tests/compiler.js:325:13)
      at Test.run (/Users/aschaar/programming/xtags-org/node_modules/nunjucks/node_modules/mocha/lib/runnable.js:213:32)
      at Runner.runTest (/Users/aschaar/programming/xtags-org/node_modules/nunjucks/node_modules/mocha/lib/runner.js:343:10)
      at /Users/aschaar/programming/xtags-org/node_modules/nunjucks/node_modules/mocha/lib/runner.js:389:12
      at next (/Users/aschaar/programming/xtags-org/node_modules/nunjucks/node_modules/mocha/lib/runner.js:269:14)
      at /Users/aschaar/programming/xtags-org/node_modules/nunjucks/node_modules/mocha/lib/runner.js:278:7
      at next (/Users/aschaar/programming/xtags-org/node_modules/nunjucks/node_modules/mocha/lib/runner.js:226:23)
      at Array.0 (/Users/aschaar/programming/xtags-org/node_modules/nunjucks/node_modules/mocha/lib/runner.js:246:5)
      at EventEmitter._tickCallback (node.js:192:40)
make: *** [test] Error 1

```

[...]

PR #570, opened 2015-10-29
Escape should not escape and mangle already escaped html entities

shoshomiga:
Consider that we are have already escaped text like:

`test &#39; &quot; &lt; &gt; &amp;`

In the browser it will look like

test &#39; &quot; &lt; &gt; &amp;

Escape will change all those & to `&amp;` and you will see

test &amp;#39; &amp;quot; &amp;lt; &amp;gt; &amp;amp;

Because the text is now

`test &amp;#39; &amp;quot; &amp;lt; &amp;gt; &amp;amp;`

Since auto escape is the default I think this behavior is bad.

I changed the regex so it doesn't do that, but its probably not the most efficient or correct way to do it.

### DIFF
--- a/src/lib.js
+++ b/src/lib.js
@@ -11,7 +11,7 @@ var escapeMap = {
     '>': '&gt;'
 };
 
-var escapeRegex = /[&"'<>]/g;
+var escapeRegex = /&(?!amp;|lt;|gt;|quot;|#39;)|["'<>]/g;
 
 var lookupEscape = function(ch) {
     return escapeMap[ch];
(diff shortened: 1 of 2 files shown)


ricordisamoa:
What if I need to display escaped HTML in pages, so I escape it twice? Won't the `safe` filter work for your variables?

shoshomiga:
That is a good question, maybe this can be a configurable option.

For my use case, the text should come already escaped from a database, but I can't be sure that it was escaped.

Hence why I think that auto escape should be used to make sure that the text is escaped.

shoshomiga:
After thinking about @ricordisamoa comment I decided that this would cause unexpected behavior and makes the escaping needlessly complicated so I'm going to close this PR

PR #587, opened 2015-11-14
Add logic to call async function in view

trquoccuong:
I add new filter is "async" and change logic of "addGlobal" .

### DIFF
--- a/src/environment.js
+++ b/src/environment.js
@@ -19,11 +19,13 @@ builtin_loaders.PrecompiledLoader = require('./precompiled-loader');
 // If the user is using the async API, *always* call it
 // asynchronously even if the template was synchronous.
 function callbackAsap(cb, err, res) {
-    asap(function() { cb(err, res); });
+    asap(function () {
+        cb(err, res);
+    });
 }
 
 var Environment = Obj.extend({
-    init: function(loaders, opts) {
+    init: function (loaders, opts) {
         // The dev flag determines the trace that'll be shown on errors.
         // If set to true, returns the full trace from the error point,
         // otherwise will return trace starting from Template.render
@@ -47,12 +49,12 @@ var Environment = Obj.extend({
 
         this.loaders = [];
 
-        if(!loaders) {
+        if (!loaders) {
             // The filesystem loader is only available server-side
-            if(builtin_loaders.FileSystemLoader) {
+            if (builtin_loaders.FileSystemLoader) {
                 this.loaders = [new builtin_loaders.FileSystemLoader('views')];
             }
-            else if(builtin_loaders.WebLoader) {
+            else if (builtin_loaders.WebLoader) {
                 this.loaders = [new builtin_loaders.WebLoader('/views')];
             }
         }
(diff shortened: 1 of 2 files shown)


carljm:
I'm sorry, but I'm afraid I don't understand the purpose or benefit of either of the changes in this pull request, so I'm closing it. If you can offer some explanation and example of why these changes are useful, I can reconsider.

The new `async` filter allows you to say `{{ func|async(cb) }}` where you could currently just say `{{ func(cb) }}` -- how is that a useful improvement?

And the change to `addGlobal` is a bit obscure, but seems to be a way to wrap function globals in a partial, but it's not clear what value that brings or how it is related to async.

trquoccuong:
I add this logic for use 
{{  func(data) | async }}. The javascript function add to view can call async( like : get data from database...). this function will have callback like filter

carljm:
I'm not sure that usage is doing what you intend. If you do `{{ func(data) | async }}` (passing no argument to the filter) the value of `cb` within the `async` filter will always be `undefined`, so why does the `async` filter even take a `cb` arg?

In any case, I don't think this belongs in nunjucks. If you want to wrap your functions in a partial-creator to delay their execution, and then later double-call them, you can do that with a utility that is separate from nunjucks.

Thanks for the contribution!

trquoccuong:
Let me explain my tricky here.
By default , nunjucks not support async function in view so I must use filter 
{{ "crazy string" | get_something}} this make my friend misunderstood. So I make a tricky here for call {{get_sidebar () | async } . Filter must have in and out, depend by in. But function only care about out. Hope nunjucks support async function soon. Sorry for my bad English I think this is only sematic problem

carljm:
No worries about English, and sorry if I am misunderstanding!

I don't really understand what you mean when you say "by default, nunjucks not support async function in view". First of all, I'm not sure what you mean by "view" -- do you mean in the template?

Can you give an actual code example of an "async function" that you would like to be able to call in a template, and that works with the code in this PR?

trquoccuong:
we build a CMS for nodejs. In cms, we support user add widgets to sidebars. In template to get all sidebar user selected, we need query database and render html and merge html. In frontend template, we call
{{ get_sidebar() | async }}. Totally, we can use filter here but we think some developer maybe misunderstand 
```
function (sidebarName, callback) {
        let app = this;

        // Find all widgets in the sidebar
        app.models.widget.findAll({
            where: {
                sidebar: sidebarName
            },
            order: ['ordering'],
            raw: true
        }).then(function (widgets) {

            // Check the sidebar has widget
            if (widgets && widgets.length) {
                let html = '';
                let resolve = Promise.resolve();

                widgets.map(function (w) {
                    // Get widget by type
                    let widget = app.widgetManager.getComponent(w.widget_type);

                    if (widget) {
                        // Get content of each widget in the sidebar
                        resolve = resolve.then(function () {
                            return widget.controllers.renderWidget(w.data)
                        }).then(function (view) {
                            return html += view;
                        }).catch(function (err) {
                            log.error(err);
                        });
                    }
                });

                // Return content of the sidebar
                resolve.then(function () {
                    callback(null, html);
                });
            } else {
                callback(null, '');
            }
        }).catch(function (err) {
            log.error(err);
        });
    }
};
```

PR #588, opened 2015-11-17
Fix #516 by exposing "dev" not in "env.opts" but "env" itself

### LINKED ISSUE #516: verbose error messages with dev mode broken
Hi,
it seems there's a bug with the ``prettifyError`` function in dev-mode. ``prettifyError`` needs the ``withInternals`` param to display verbose error messages [1]. If compiling a template fails, ``prettifyError`` is called with the ``dev``-flag in the ``env``-object [2].

The problem is now that setting the ``dev``-flag (with ``nunjucks.configure({dev: true})``) changes ``this.env.opts.dev``, so a lookup for the flag is done at the wrong object and the dev-mode is never activated.

[1] ``exports.prettifyError = function(path, withInternals, err) {``
[2] ``var err = lib.prettifyError(this.path, this.env.dev, _err);``

soswow:
because it is used like this + a test. Not sure about tests. Maybe there is more elegant way to test that. I just took previous test and used it's stacktrace

### DIFF
--- a/src/environment.js
+++ b/src/environment.js
@@ -31,7 +31,7 @@ var Environment = Obj.extend({
         //  the library)
         // defaults to false
         opts = this.opts = opts || {};
-        this.opts.dev = !!opts.dev;
+        this.dev = !!opts.dev;
 
         // The autoescape flag sets global autoescaping. If true,
         // every string variable will be escaped by default.
--- a/tests/compiler.js
+++ b/tests/compiler.js
@@ -755,6 +755,19 @@
             finish(done);
         });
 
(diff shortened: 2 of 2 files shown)


soswow:
Not sure why those checks are failing =\

carljm:
I think the failing checks were due to a transient issue with a broken version of jshint. More recent Travis builds on master have passed.

The test here looks fine to me, but on a closer look I think the fix is the reverse of what I'd prefer. All the other options are stored in `env.opts`, I'd rather keep consistency there and fix the usage.

carljm:
Fixed in 0b2fb96ae27e9ad34d32e7b1786cbc6ece190603

Issue #359, opened 2015-01-25
Why is there no `striptags` filter?


philipwalton:
I'm just curious because it's in Jinja2, and it seems like there's no good reason not to include it.

zacwitte:
I was wondering the same with some other filters like `map` and `split` which are super easy to add as a custom filter, but that adds other complexity

jlongster:
I'm open to a PR if you get around to it before me!

jbmoelker:
This issue is resolved in #589. Result on [Templating docs > Strip tags](https://mozilla.github.io/nunjucks/templating.html#striptags-value-preserve_linebreaks). 
So this issue should be closed.

PR #589, opened 2015-11-18
Add striptags filter

antogyn:
I migrated from swig to nunjucks and realized there was no striptags filter. There is already an opened issue : #359

One of the tests is very similar to [this one (from jinja)](https://github.com/mitsuhiko/jinja2/blob/master/tests/test_filters.py#L68-L73)

### DIFF
--- a/docs/cn/templating.md
+++ b/docs/cn/templating.md
@@ -634,6 +634,7 @@ Nunjucks 已经支持了大部分 jinja 的过滤器 (点击查看文档)。
 * [slice](http://jinja.pocoo.org/docs/templates/#slice)
 * [sort](http://jinja.pocoo.org/docs/templates/#sort)
 * [string](http://jinja.pocoo.org/docs/templates/#string)
+* [striptags](http://jinja.pocoo.org/docs/templates/#striptags)
 * [title](http://jinja.pocoo.org/docs/templates/#title)
 * [trim](http://jinja.pocoo.org/docs/templates/#trim)
 * [truncate](http://jinja.pocoo.org/docs/templates/#truncate)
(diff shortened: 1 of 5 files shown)


carljm:
Looks like the build is failing. I'd be happy to merge this with a green build.

antogyn:
The build is also failing on master (of this repo) with the same messages on my local machine with node 0.10 and 0.11. I don't think it comes from this PR

antogyn:
There seems to be a problem with the last version of jshint. Using jshint 2.8.x fixes the problem on both versions. Should I make a second PR to fix this ?

antogyn:
Done.

ivan-kleshnin:
Not a very good implementation because it strips whitespace.
You want to pipe `striptags`, `nl2br`, `safe` in most cases when handle custom messages from user.
But `nl2br` can't help here because `striptags` just killed all whitespace.

carljm:
Can you file a pull request?

ivan-kleshnin:
Yes, I'll give it a second view and make a PR.

carljm:
Thank you!

antogyn:
Keep in mind that what you propose is different from jinja's implementation and won't pass the same test jinja has.

ivan-kleshnin:
Yes, I'll update tests accordingly.

carljm:
Hmmm. Thanks for pointing that out, @antogyn. I think I prefer to maintain Jinja2 compatibility here by default, but I'm not opposed to adding this as an optional behavior controlled by a flag, e.g. `preserve_whitespace=true` (defaulting to `false`).

ivan-kleshnin:
Yes, I was going to put it under non-default flag. Will try to send PR tomorrow.

ivan-kleshnin:
Sorry for delay. A tons of work activity.
Recently sent this: https://github.com/mozilla/nunjucks/pull/619
I named it `preserve_linebreaks` because it rather normalizes whitespace (and linebreaks...).

Issue #531, opened 2015-09-28
nunjucks.render always uses config from last call to nunjucks.configure


andrewchernow:
I am using v2.1.0 from the browser only.

`nunjucks.render()` is a little bit dangerous. Since `nunjucks.configure` caches the `Environment` it creates, a second call to `configure` will cause an innocent victim's `Environment` to be overwritten; if they choose to use `nunjucks.render` rather than `Environment.render`. 

I couldn't find this documented anywhere. I think it should be (warning) or `nunjucks.render` should require an `Environment` argument.

I'm guessing it is coded this way since the most common use case is to create just one `Environment` or not create one at all and just start calling `nunjucks.render`.

carljm:
Yeah. I guess it was done this way so that you could just call `nunjucks.configure` (ignoring the returned environment) and later `nunjucks.render`, and have it respect the configuration you'd set. But I agree this is a bunch of implicit global state that is problematic.

In the short term we should just document clearly that `nunjucks.render` always uses the configuration from the last call to `nunjucks.configure`, so you probably don't want to ever use it if you are creating more than one environment.

In the longer run, I think the best/simplest option might be just to deprecate `nunjucks.render` entirely, and require always creating an environment and calling `env.render`.

andrewchernow:
Unless I am missing something, can't you just call `nunjucks.render` without configure, which would use all of the default options? I am only using this from the browser so I am not sure what express would do.

I agree that deprecating `nunjucks.render` is probably the way to go.

carljm:
@andrewchernow 

> Unless I am missing something, can't you just call nunjucks.render without configure, which would use all of the default options?

Yes, you can, and obviously that's one use case for `nunjucks.render`. What I was trying to explain above is why there is interaction between `nunjucks.configure` and `nunjucks.render` at all (instead of `nunjucks.render` just caching its own default environment). The latter was my first thought of how to fix this (remove the caching from `nunjucks.configure` and just have `nunjucks.render` cache its own environment), but then I realized that would break the use case I described above.

> Absolutely terrific engine BTW. nice job.

Credit to @jlongster for that - I just help with maintenance.

carljm:
Added a documentation note in c6eee65

carljm:
[...]

PR #593, opened 2015-11-24
Doc interaction between nunjucks.configure and the simple API (french)

forresst:
Translation in french 
ref #531


carljm:
Thanks again! Is there anything I can do to make this process easier for you? I'd be happy to give you notice before I make a release so that you can review the changes since the last release and do them all at once? Or if you prefer to just follow commits and do it right away that's fine too!

forresst:
@carljm It's very nice to want to help me. Do not worry ! I follow the commits as soon as I have some time. However, do not hesitate to notify me if a translation is to be done. I'll try to do it as quickly as possible.

PR #606, opened 2015-12-10
Build browser bundles with a UMD wrapper.

jgerigmeyer:
Fixes https://github.com/mozilla/nunjucks/issues/507, replacing https://github.com/mozilla/nunjucks/pull/556.

Note that this should be considered a breaking change, at least for people who were using the ``exports-loader`` to expose the ``nunjucks`` var before.

### DIFF
--- a/bin/bundle.js
+++ b/bin/bundle.js
@@ -35,7 +35,8 @@ var config = {
     output: {
         path: path.join(__dirname, '../browser'),
         filename: TARGET,
-        library: 'nunjucks'
+        library: 'nunjucks',
+        libraryTarget: 'umd'
     },
     node: {
         process: 'empty'


carljm:
@jgerigmeyer How will it break for those people? What documentation can we provide for them on how to make the upgrade? Is there any feasible way we can make the change opt-in, or make it work both ways, and not have it be a breaking change?

jgerigmeyer:
I'm not a webpack expert, but as far as I can tell there's no way to bundle nunjucks using both ``var`` and ``umd`` targets. If we switch to ``umd`` (i.e. this PR), users who are using the ``exports-loader`` to expose the ``nunjucks`` var will get a ``Uncaught ReferenceError: nunjucks is not defined``.

People loading it via a standard ``script`` tag should still be able to use it fine (it is still exposed on ``window`` in a browser context), and people loading it via AMD or CommonJS bundlers will now simply be able to ``require('nunjucks');``.

As far as opt-in, the only option I see would be adding a new ``nunjucks-umd`` bundle or something like that, but I'm not sure that's much better than the current situation. Either way, people using bundlers/compilers need to add special configuration.

Do others have thoughts? Am I missing something?

Rowno:
I think bumping the major version and adding a comment in the release notes is fine. It should be an easy change for people using the `exports-loader` workaround right?

at0g:
Won't it still fail in server environments (window.nunjucksPrecompiled for example)?
UMD should work in browser, AMD, and server environments.

jgerigmeyer:
@at0g I'm not sure I understand your question; can you clarify (with examples as necessary)?

I don't see how this PR changes how nunjucks can be used in server environments.

at0g:
@jgerigmeyer my understanding of [UMD](https://github.com/umdjs/umd) is that it can be used anywhere, not just the browser.
Currently the `browser/nunjucks*` files are browser only - trying to run these in a server environment will fail. It would be great to have a true UMD bundle, but I think there is a lot more involved than just setting the library target.

jgerigmeyer:
Ah, I see. Yes, you're correct. In this case I prefer ``umd`` over ``commonjs`` because it supports native browser ``script`` tags as well as AMD/CommonJS module bundlers (e.g. webpack and browserify). But I haven't played with (or honestly put a lot of thought into) the server-only use-case.

jgerigmeyer:
I guess my impression was that nunjucks can already be used on the server by ``require``ing the module instead of using one of the browser bundles. Is there a particular use-case that doesn't work which you'd like nunjucks to support? If so, maybe file a new bug (and, even better, submit a PR)? :)

at0g:
I created [this branch](https://github.com/at0g/nunjucks/tree/feature/umd-precompiled) in a fork yesterday that addresses some (but not all) of the steps to create a UMD bundle, but there are still some hairy bits (such as relative templates).

One good reason (IMO) to have UMD support, is that any project using nunjucks cannot be published as a UMD library, as it needs to switch which version of nunjucks is being used (browser or server).

at0g:
I'm happy to open an issue/PR from the branch above and move the conversion there if you like?

jgerigmeyer:
Yeah, that makes a lot of sense. I'd say open a new issue/PR, as this PR still solves a real problem that current (browser) users are having.

at0g:
ok done

carljm:
See my latest comment on #507 -- I think this PR would fix the immediate pain point people are having, but I'm not convinced it's really the right fix. Closing for now, can reopen and merge if we decide that I'm wrong :-)

DylanPiercey:
A UMD export would be awesome. +1

PR #607, opened 2015-12-12
chokidar on demand

### LINKED ISSUE #600: Build without chokidar
I believe `chokidar` is in use for watching purposes only, and i don't think it is required by everyone. While building a project i will have a file watcher which looks for all source files and then restart the server on change. 

This minimizes the need of having views only watcher, and chokidar will install it's own dependencies where everything is useless for me

at0g:
- make chokidar an optional dependency
- only require chokidar when `{ watch: true }`
- if chokidar is not installed and `{ watch: true }`, notify the user with a console error (but don't fail)

Fixes #600

### DIFF
--- a/src/node-loaders.js
+++ b/src/node-loaders.js
@@ -4,7 +4,6 @@ var fs = require('fs');
 var path = require('path');
 var lib = require('./lib');
 var Loader = require('./loader');
-var chokidar = require('chokidar');
 var PrecompiledLoader = require('./precompiled-loader.js');
 
 // Node <0.7.1 compatibility
@@ -34,8 +33,8 @@ var FileSystemLoader = Loader.extend({
         }
 
         if(opts.watch) {
-            // Watch all the templates in the paths and fire an event when
-            // they change
+            var chokidar = require('chokidar');
+            // Watch all the templates in the paths and fire an event when they change
             var paths = this.searchPaths.filter(function(p) { return existsSync(p); });
             var watcher = chokidar.watch(paths);
             var _this = this;
--- a/docs/api.md
+++ b/docs/api.md
@@ -86,6 +86,7 @@ directory, and the following options are available in **opts**:
 * **trimBlocks** *(default: false)* automatically remove trailing newlines from a block/tag
 * **lstripBlocks** *(default: false)* automatically remove leading whitespace from a block/tag
 * **watch** *(default: false)* reload templates when they are changed (server-side)
+    If using watch, ensure chokidar is installed first: `$ npm i chokidar`
 * **noCache** *(default: false)* never use a cache and recompile templates each time (server-side)
 * **web** an object for configuring loading templates in the browser:
   * **useCache** *(default: false)* will enable cache and templates will never see updates.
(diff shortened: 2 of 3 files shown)


jgerigmeyer:
:+1:

src/node-loaders.js
```diff
@@ -34,20 +33,26 @@ var FileSystemLoader = Loader.extend({
...
+                    fullname = path.resolve(fullname);
+                    if(event === 'change' && fullname in _this.pathsToNames) {
+                        _this.emit('update', _this.pathsToNames[fullname]);
+                    }
+                });
+                watcher.on('error', function(error) {
+                    console.log('Watcher error: ' + error);
+                });
+            }
+            catch (err) {
```
carljm: This seems like a really broad catch, which could easily catch some other (unexpected) error, and proceed to give a wrong and misleading message on the console.

Where exactly will an error be raised if chokidar isn't installed? On the `require('chokidar')` line, right? So why not just wrap that line only?

carljm:
This could use a note in the documentation for `watch: true`? Other than that and the one code comment I just made, this looks merge-ready to me.

FWIW, personally I see this as the first step on a path to eventually deprecating and removing `watch: true` entirely. I think it's out of scope for a templating language to have its own watch system; that should be left in the realm of gulp et al.

carljm:
Even with the outpouring of positive response to #610, I'm still planning to merge this -- it'll help for future releases in the 2.x series, since it's backward-compatible, whereas I can't remove `watch: true` entirely until 3.0.

atian25:
+1 for this.  when will this merge && publish ?

any plan and roadmap for 3.x?

carljm:
@atian25 if you file a PR based on the work here that addresses my comments above, I will merge it!  Otherwise I'll get to it when I can, but I already spent quite a few hours on nunjucks this week, so that might not be right away.

carljm:
Re 3.x roadmap, I'm beginning to develop one. So far it involves more removals than additions :) mostly it depends on what people are motivated enough to file PRs for, I know I won't have time to personally fix nearly all the open issues.

atian25:
@at0g any plan at this?

at0g:
Hello, sorry about the slow reply, been a busy time of year.

I've updated the PR with your feedback:
- Narrow the catch block @jgerigmeyer 
- Added a note to the docs to install chokidar when setting `watch: true` @carljm 

Let me know if there's anything else, otherwise happy new year!

ricordisamoa:
If chokidar is not installed and `{ watch: true }`, why not fail fast?

carljm:
@at0g This looks great! Unfortunately the linter is complaining; I think it doesn't like the fact that you have the declaration of `var chokidar` inside that `try` block, while the usage of it is outside. Can you just stick a `var chokidar` up one level instead?

@ricordisamoa not sure I understand your comment - this PR does fail as fast as is possible if chokidar is not installed.

ricordisamoa:
@carljm Oh sorry, I meant throwing an error instead of logging.

carljm:
@ricordisamoa Oh! I actually missed that, thanks for pointing it out. I don't feel strongly, but yeah I think we may as well just throw in that case. There's no reason to proceed with an invalid configuration, the user should just fix it one way or another (install chokidar, or don't set `watch: true`).

TrySound:
Any news?

atian25:
@carljm @at0g  any news about this?

Issue #608, opened 2015-12-13
Provide true UMD support for nunjucks-slim


at0g:
It would be preferable if there was a true UMD build of nunjucks that can render templates in both the browser, AMD, or server environments from a single file.

My use case is that as [an author of a package that depends on nunjucks](https://www.npmjs.com/package/nunjucks-loader), I cannot bundle my library with UMD support as different versions of nunjucks must be used for both the browser and server. 

Given that the UMD wrapper is a packaging step, it would be acceptable to only have precompiled / slim support.

Any features that are not supported in both environments would be dropped (lowest common denominator) - for example, relative template paths are not supported in the browser, so would not be supported in a UMD build.

carljm:
Closing in favor of #609 to keep the conversation in one place -- we don't need both the issue and the PR.

PR #609, opened 2015-12-13
UMD precompiled

at0g:
**THIS IS NOT READY TO MERGE**

#608 

Convert precompiled templates to use a UMD wrapper instead of `window.nunjucksPrecompiled`, this is a step towards making a true UMD compatible nunjucks-slim version:

- In browser, compiled templates are attached to `window.nunjucksPrecompiled`
- In an AMD environment, `nunjucksPrecompiled` is a named module
- in node, compiled templates are attached to `GLOBAL.nunjucksPrecompiled`

The generated browser files are not included in this commit to keep the source changes obvious - `npm run browserfiles` to generate them.

### DIFF
--- a/bin/bundle.js
+++ b/bin/bundle.js
@@ -35,7 +35,8 @@ var config = {
     output: {
         path: path.join(__dirname, '../browser'),
         filename: TARGET,
-        library: 'nunjucks'
+        library: 'nunjucks',
+        libraryTarget: 'umd'
     },
     node: {
         process: 'empty'
--- a/src/environment.js
+++ b/src/environment.js
@@ -11,6 +11,9 @@ var runtime = require('./runtime');
 var globals = require('./globals');
 var Frame = runtime.Frame;
 var Template;
+// Set the root for global variables (precompiled templates export to this)
+var root = typeof window === 'object' ? window : GLOBAL;
+
 
 // Unconditionally load in this loader, even if no other ones are
 // included (possible in the slim browser build)
@@ -63,9 +66,9 @@ var Environment = Obj.extend({
         // It's easy to use precompiled templates: just include them
         // before you configure nunjucks and this will automatically
         // pick it up and use it
-        if(process.env.IS_BROWSER && window.nunjucksPrecompiled) {
+        if(root.nunjucksPrecompiled) {
             this.loaders.unshift(
-                new builtin_loaders.PrecompiledLoader(window.nunjucksPrecompiled)
+                new builtin_loaders.PrecompiledLoader(root.nunjucksPrecompiled)
             );
         }
 
--- a/src/precompile-global.js
+++ b/src/precompile-global.js
@@ -4,20 +4,52 @@ function precompileGlobal(templates, opts) {
     var out = '', name, template;
     opts = opts || {};
 
-    for ( var i = 0; i < templates.length; i++ ) {
-        name = JSON.stringify(templates[i].name);
-        template = templates[i].template;
+    out += '// nunjucks precompiled templates\n';
+    out += '(function (root, factory) {\n' +
+        '\tif (typeof define === \'function\' && define.amd) {\n' +
+            '\t\t// AMD. Register as a named module.\n' +
+            '\t\tdefine(\'nunjucksPrecompiled\', [], function () {\n' +
+                '\t\t\treturn (root.nunjucksPrecompiled = factory());\n' +
+            '\t\t});\n' +
+        '\t} else if (typeof module === \'object\' && module.exports) {\n' +
+            '\t\t// Node. Does not work with strict CommonJS,\n' +
+            '\t\t// but only CommonJS-like environments that support module.exports, like Node.\n' +
+            '\t\tGLOBAL.nunjucksPrecompiled = factory();\n' +
+        '\t} else {\n' +
+            '\t\t// Browser globals\n' +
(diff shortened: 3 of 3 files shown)


rhengles:
This was already discussed before: #302 #337 #418

You'll need to create a external module such as these:

https://github.com/rhengles/nunjucks-cjs
https://github.com/rhengles/nunjucks-amd
https://github.com/rhengles/nunjucks-esm

carljm:
@at0g I'm inclined to go with the already-in-place solution that's more generic and flexible, even though it requires you to provide a wrapper package. Is there some way in which the existing precompilation hooks don't let you do what you need to do?

carljm:
Tentatively closing pending further feedback.

carljm:
It's certainly an option to create a `nunjucks-umd` project -- if there are changes needed to nunjucks itself to make that possible, I'm more interested in that than in building in support for Yet Another Build.

at0g:
@carljm - What I was trying to do with this PR is basically remove the `window` dependency, as it's problematic outside of anything but the browser.

In terms of changes needed to nunjucks, my use case is for nunjucks-loader (a webpack loader plugin).
webpack offers a `output.libraryTarget` option where `umd` is a valid target. Because nunjucks is not umd compatible, it makes it difficult to meet that target (as of nunjucks 2.x I have dropped non-browser support). A UMD slim build would be ideal.

Even if I was to use the `wrapper` option on precompile as suggested by @rhengles, I think it's flawed because:

1. It does not solve the problem outside the browser with [this line of code](https://github.com/mozilla/nunjucks/blob/master/src/environment.js#L66).

2. `asFunction` is broken if templates have dependencies - see comment on #302 -although I got around this by [rewriting the templates with webpacks require](https://github.com/at0g/nunjucks-loader/blob/master/index.js#L116)
> Yes, it will always also put it on nunjucksPrecompiled regardless. That's a bit of a hack, but it's needed for template loading. If you ever use include, import, etc. the system needs a registry of all the templates. So you could use this and put each template in a module, and load those modules separately, but you would also need to load any templates that you include in another template manually, then you could use include.

That said, IMO having to load precompiled templates before nunjucks is the wrong order of dependencies anyway - the nunjucks runtime is a dependency of the precompiled templates - not the other way around. I believe this is really what the `window.nunjucksPrecompiled` global is facilitating.

I would be great if this worked (psuedo code):

```
1. load nunjucks (no deps, exports nunjucks)
2. load precompiled templates (depends on nunjucks, exports hash of templates)
3. precompiled templates have a UMD wrapper and initialise with a call like this:
exports[templateName] = compiledTemplate;
var env = nunjucks.getEnvironment() || new nunjucks.Environment();
env.addLoader(new nunjucks.loaders.PrecompiledLoader(exports));
```

^ Maybe food for thought on the 3.x release.

carljm:
@at0g I'm not sure I fully understand why the `wrapper` option doesn't work for you. To take your numbered list in order:

1. Yes, that line of code is browser-specific, but it's also nothing more than an almost-superfluous convenience for the common browser case, so you don't have to explicitly specify that you want to use the `PrecompiledLoader` -- it just detects `window.nunjucksPrecompiled` and adds the loader automatically. We could remove that line of code entirely and nothing fundamental would change, you'd just have to add the `PrecompiledLoader` yourself instead of nunjucks doing it for you. Your desired pseudocode explicitly specifies the loader anyway, so this convenience is irrelevant to you; just pretend it doesn't exist.

2. `asFunction` is not relevant to the `wrapper` approach. It seems like maybe you got confused by some of the back-and-forth between @jlongster and @rhengles on #302 (or else I'm confused!)? The critique of `asFunction` that you quote is @rhengles explaining why the older `asFunction` wasn't sufficient and the new `wrapper` option was needed instead -- and @jlongster was convinced by that argument and merged the new `wrapper` option. `asFunction` is a feature solely of the default "global" wrapper shipped with nunjucks, which populates `window.nunjucksPrecompiled` and respects the `asFunction` option. If you write your own wrapper, you are no longer using the default global wrapper, so `window.nunjucksPrecompiled` will not be populated, and `asFunction` is no longer relevant at all. And you can wrap the precompiled templates in whatever module packaging you want. See the sample wrappers @rhengles linked. (One small problem here is that the `wrapper` option is not documented -- so, @rhengles - thanks for your awesome work on #302, do you feel like filing a follow-up PR to add documentation for it? :P)

I think that you can make your pseudocode work today by using the custom-wrapper approach. Your custom wrapper can put the UMD boilerplate around the precompiled-templates hash, just like you want, and the `PrecompiledLoader` already works the way you want (accepts a precompiled templates hash on initialization).

I'm not _totally_ closed to the idea that a UMD wrapper should replace the current global wrapper as the default precompilation wrapper for nunjucks 3.x, but I'd like to see it built and working first as an external package and play with it on some real projects.

Please let me know if I'm missing something important that's actually preventing you from implementing your pseudocode today using a custom UMD wrapper.

at0g:
Thanks for the detailed answers.

I will try an amd wrapper option to implement the psudocode in a separate
project. If/when I get it running I'll let you know.

Cheers

carljm:
Excellent! And if you run into roadblocks that seem like they need to be fixed on the nunjucks side, don't hesitate to ping about it.

rhengles:
> so, @rhengles - thanks for your awesome work on #302, do you feel like filing a follow-up PR to add documentation for it? :P
@carljm Sure, I'll try to do it in the holidays. :D

PR #621, opened 2015-12-25
docs for 'wrapper' option of the precompile API

rhengles:
As requested on #609, this PR adds docs for the 'wrapper' option of nunjucks' precompile API introduced in #302, and the corresponding option in the CLI in #418.

Please let me know if there's any improvement I can make.


carljm:
Looks great, thank you!

PR #623, opened 2015-12-27
filter.escape should not escape SafeString, close #618

### LINKED ISSUE #618: filter.escape should not escape SafeString
https://github.com/mozilla/nunjucks/blob/master/src%2Ffilters.js#L118

```
    escape: function(str) {
        if(typeof str === 'string' ||
           str instanceof r.SafeString) {
            return lib.escape(str);
        }
        return str;
    },
```

atian25:


atian25:
rebase done.

src/filters.js
```diff
@@ -117,7 +117,7 @@ var filters = {
 
     escape: function(str) {
         if(typeof str === 'string' ||
-           str instanceof r.SafeString) {
+           str instanceof r.SafeString === false) {
```
carljm: This definitely doesn't look right; it would cause this filter to attempt to escape any object which is not a `SafeString`. I think instead the entire conditional can be simplified to just `if(typeof str === 'string')` removing the `||` entirely.

carljm:
Thanks for the PR!

Normally we don't compile the browser versions for every single change -- that just dramatically increases the likelihood of conflicts between pull requests. Can you update this PR to just add the test and change the source, but not include the compiled files?

Also, please add a note in the changelog about this change. It's backwards-incompatible, so I think it'll need to go into v3.

atian25:
@carljm thanks for review, had update.

carljm:
Merged, thanks!

PR #628, opened 2016-01-05
Try to fix browser tests, close#612

oyyd:
> Due to attempt to modify read-only property length -- see http://mozilla.github.io/nunjucks/files/tests/browser/

It's the problem of expect.js so that I have updated the newest version in `/tests/browser`.

I also updated the newset version of mocha in `/tests/browser` as it will cause a timeout failure in browsers.

But there are still two failing tests related to `ignore missing` which was not tested in browsers before and I don know how to do with them.

### DIFF
--- a/index.js
+++ b/index.js
@@ -20,6 +20,7 @@ module.exports.parser = require('./src/parser');
 module.exports.lexer = require('./src/lexer');
 module.exports.runtime = require('./src/runtime');
 module.exports.lib = lib;
+module.exports.nodes = require('./src/nodes');
 
 module.exports.installJinjaCompat = require('./src/jinja-compat.js');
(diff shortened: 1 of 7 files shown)


carljm:
Thank you! This is a great pull request, much appreciated.

The `ignore missing` tests are failing because the `WebLoader` throws an error instead of returning `null` on a missing template. I'll merge this and then fix that separately.

carljm:
FYI I fixed the WebLoader issue and pushed v2.3.0, and the online docs now have passing tests: http://mozilla.github.io/nunjucks/files/tests/browser/

Thanks again!

oyyd:
@carljm :+1: Good job

PR #629, opened 2016-01-06
Add "sum" filter

lazopm:
http://jinja.pocoo.org/docs/dev/templates/#sum

### DIFF
--- a/src/filters.js
+++ b/src/filters.js
@@ -358,7 +358,25 @@ var filters = {
 
         return res;
     },
-
+    
+    sum: function(arr, start, attr) {
+        if(typeof start !== 'number'){
+            start = 0;
+        }
+        
+        if(attr) {
+            arr = lib.map(arr, function(v) {
+                return v[attr];
+            });
+        }
+        
+        for(var i = 0; i < arr.length; i++) {
+            sum += arr[i];
+        }
+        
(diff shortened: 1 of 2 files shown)


src/filters.js
```diff
@@ -358,7 +358,25 @@ var filters = {
 
         return res;
     },
-
+    
+    sum: function(arr, start, attr) {
```
carljm: looks like these arguments are in the opposite order from the jinja version

src/filters.js
```diff
@@ -358,7 +358,25 @@ var filters = {
...
+        }
+        
+        if(attr) {
+            arr = lib.map(arr, function(v) {
+                return v[attr];
+            });
+        }
+        
+        for(var i = 0; i < arr.length; i++) {
+            sum += arr[i];
```
carljm: This var seems to be undeclared. And also you aren't using the `start` value anywhere. Were those intended to be the same var?

carljm:
Made a couple comments, and tests are failing (due to the same issue pointed out in my comments). But with the comments addressed, this looks good, thanks!

lazopm:
@carljm Does the tests look OK?

carljm:
Looks good, thanks! Merging.

Issue #626, opened 2016-01-04
Cannot catch error, thrown from global function


KSDaemon:
Hi! 
I'm trying to cover with tests (mocha/expect) some code, that register global functions in nunjucks, which throw errors in some invalid cases and cannot catch them in expects.
I don't know, is it a valid by design?
I've made a test for this case:
```javascript
        it('should return errors from globals', function(done) {
            var env = new Environment(new Loader(templatesPath));

            env.addGlobal('err', function() {
                throw new Error('Global error');
            });

            expect(equal('{{ err() }}', 'does not matter', env)).to.throwError();

            finish(done);
        });
```
that results in:
```
  1) global should return errors from globals:
     Template render error: (unknown path)
  Error: Global error
      at Object.exports.prettifyError (src/lib.js:9:1149)
      at src/environment.js:9:23995
      at root [as rootRenderFunc] (eval at <anonymous> (src/environment.js:9:27378), <anonymous>:16:3)
      at Obj.extend.render (src/environment.js:9:23652)
      at render (tests/util.js:98:22)
      at equal (tests/util.js:36:19)
      at Context.<anonymous> (tests/globals.js:155:20)
```
But error was not catched :(

Any thoughts about this? May be i'm wrong at some point?

carljm:
I'm not sure what's happening here exactly, but I think you should start by simplifying your test. For instance, you are using the `equal` utility function, which internally does a string equivalence check (using `.to.be(...)`), which is useless for this test, and I don't know how that might impact Mocha's error-catching. Just use `render` directly instead.

Until we have a simplified reproduction case showing that this is a problem with nunjucks (as opposed to just an error in Mocha usage), I'm closing this issue.

KSDaemon:
Hi, @carljm ! I've added a test for this case. Hope it is just what you are asking for.

carljm:
Yes, that's the simplification of the test case I asked for. Thanks!

I'm not sure what's happening here; nunjucks does catch and re-throw errors from template rendering, but I don't know why the re-thrown error wouldn't still satisfy Mocha.

I'm also not convinced this necessarily needs to be fixed. If I wanted to write a test covering edge-case behavior of a filter or global function, I'd just have the test call the filter/function directly, instead of through rendering a template.

I'll reopen this, though -- if you can figure out what's happening and find a fix that doesn't have other negative effects, I'm open to looking at it.

carljm:
[...]

PR #630, opened 2016-01-07
Added test for covering error handling in globals

KSDaemon:
Hi! This test shows, that error, thrown from global, can not be caught by test suite (mocha).
In console, we can see that error was happened, but the upper one is from nunjucks itself.
```
     Template render error: (unknown path)
  Error: Global error
      at Object.exports.prettifyError (src/lib.js:9:1149)
      at src/environment.js:9:23995
      at root [as rootRenderFunc] (eval at <anonymous> (src/environment.js:9:27378), <anonymous>:16:3)
      at Obj.extend.render (src/environment.js:9:23652)
      at render (tests/util.js:98:22)
      at Context.<anonymous> (tests/globals.js:156:20)
```
Relatedto #626


carljm:
I guess now this just adds a passing test without changing behavior? That's useful, I'll go ahead and merge.

PR #631, opened 2016-01-08
Fix unexpected whitespace control behavior, close #595.

### LINKED ISSUE #595: Dash tags -%} remove space between tags in an unexpected way
``` 
{% macro dash(a) -%} {{ a }} {{ a }} {%- endmacro %}
/{{ dash("_") }}/
``` 
Produce /_ _/ as expected (no spaces outside, one inside)
But : 
``` 
{% macro dash(a) -%}{{ a }} {{ a }}{%- endmacro %}
/{{ dash("_") }}/
``` 
Produce /__/ (no space outside, an no inside)

It that the expected behaviour ? "If the tag with tash is not followed by a space, then stip spaces after the next one"

oyyd:
I followed jinja2 to fix #595.

### DIFF
--- a/src/parser.js
+++ b/src/parser.js
@@ -1202,6 +1202,7 @@ var Parser = Object.extend({
                                                                   data)]));
             }
             else if(tok.type === lexer.TOKEN_BLOCK_START) {
+                this.dropLeadingWhitespace = false;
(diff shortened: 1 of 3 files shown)


carljm:
Looks great, thank you!

PR #632, opened 2016-01-10
Support whitespace control in comment blocks, close #332

### LINKED ISSUE #332: Comments render new lines
The following:
```
{# A useful comment #}
Foo
```
renders to:
```

Foo
```

Is there a way to prevent the new line from happening?

(I've tried `-#}`...)

oyyd:
Seems like jingja2 also supports whitespace control in variable blocks. Should nunjucks support that too?

`trimBlocks` and `lstripBlocks` only works in statement blocks(`{% %}`). Maybe They also should be supported in other blocks.

### DIFF
--- a/src/parser.js
+++ b/src/parser.js
@@ -1189,7 +1189,7 @@ var Parser = Object.extend({
 
                 // Same for the succeding block start token
                 if(nextToken &&
-                   ((nextToken.type === lexer.TOKEN_BLOCK_START &&
+                    ((nextToken.type === lexer.TOKEN_BLOCK_START &&
                       nextVal.charAt(nextVal.length - 1) === '-') ||
                     (nextToken.type === lexer.TOKEN_COMMENT &&
                       nextVal.charAt(this.tokens.tags.COMMENT_START.length)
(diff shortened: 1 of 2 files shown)


src/parser.js
```diff
@@ -1189,8 +1189,11 @@ var Parser = Object.extend({
 
                 // Same for the succeding block start token
                 if(nextToken &&
-                   nextToken.type === lexer.TOKEN_BLOCK_START &&
-                   nextVal.charAt(nextVal.length - 1) === '-') {
+                    ((nextToken.type === lexer.TOKEN_BLOCK_START &&
+                      nextVal.charAt(nextVal.length - 1) === '-') ||
+                    (nextToken.type === lexer.TOKEN_COMMENT &&
+                      nextVal.charAt(this.tokens.tags.COMMENT_START.length)
```
carljm: Why did you choose to use `this.tokens.tags.COMMENT_START.length` here instead of `nextVal.length - 1` as used in the existing clause?

carljm:
Looks good, thanks! I had one question in review, but I don't need an answer before merging.

And yes, I would prefer to match Jinja in supporting whitespace control for variable substitutions as well.

src/parser.js
oyyd: Because the `TOKEN_COMMENT` contains `COMMENT_START` and `COMMENT_END`. It's a whole token so that the `nextVal` here is not `{%`.  

Maybe breaking the `TOKEN_COMMENT` into three parts would be better. But It need change more code so that I use the tag instead.

oyyd:
Matching jinja would be nice :+1: .

PR #634, opened 2016-01-11
fix #571, make asyncif handle error

### LINKED ISSUE #571: Template error supressed when it's included inside {% if %}
I have a template with an error, i.e. like `bad.html`:

     {% fail %}some output

When I include this file I get a proper error message:

    {% include './bad.html' %}

    Template render error: (backend/src/views/test/include.html)
     Template render error: (backend/src/views/test/bad.html) [Line 1, Column 2]
     unknown block tag: fail

However when I include this file inside an `{% if %}` tag I get no error:

    {% if true %}
    {% include './bad.html' %}
    {% endif %}

The included template just stops rendering at the point of failure, I see neither `some output` nor error message.

oyyd:
Fix #571.

### DIFF
--- a/src/compiler.js
+++ b/src/compiler.js
@@ -581,7 +581,7 @@ var Compiler = Object.extend({
     compileIfAsync: function(node, frame) {
         this.emit('(function(cb) {');
         this.compileIf(node, frame, true);
-        this.emit('})(function() {');
+        this.emit('})(' + this.makeCallback());
         this.addScopeLevel();
     },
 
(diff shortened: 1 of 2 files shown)


carljm:
Looks great, thanks!

cristim:
We are still having this issue, and since there were no releases since this was merged, we still have this error in production.

Can anyone provide an ETA for when we can have this fix released?

carljm:
ETA for a release is "anytime now, as soon as I get around to it." I'll try to get to it today.

Issue #596, opened 2015-11-28
urlize filter dissolves meaningful whitespace


curiousdannii:
The urlize filter splits by whitespace and then joins with " ", which means any line breaks, tabs etc will be lost.

I can't see why it couldn't instead do something like this, where the replace function is the one currently being passed to map():

```
words = str.replace( /\S+/g, function() )
```

carljm:
The goal for the built-in filters is Jinja2 compatibility, and the Jinja2 `urlize` filter does not collapse whitespace, so this is indeed a bug. Thanks for the report!

pbu88:
I think this issue can be closed now :)

carljm:
Yep, thanks for pointing it out.

PR #637, opened 2016-01-18
Makes urlize conserve any white space char

pbu88:
Addresses issue #596.

The fix was rather simpler, I took the same approach that jinja2 `urlize` [uses](https://github.com/mitsuhiko/jinja2/blob/b78f2617b04cc549ed5fea4eca944c4d8054d187/jinja2/utils.py#L19) and is relying on the feature of `split` with regular expressions: (from [docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/split#Description))

> If separator is a regular expression that contains capturing parentheses, then each time separator is matched, the results (including any undefined results) of the capturing parentheses are spliced into the output array. However, not all browsers support this capability.

Hope this helps!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ Changelog
 master (unreleased)
 -------------------
 
+* Fix a bug in urlize that would remove line break, tabs and any other
+  "white spacey" char different than white space.
 * Don't suppress errors inside {% if %} tags. Thanks Artemy Tregubenko for
   report and test, Ouyang Yadong for fix. Merge of
   [#634](https://github.com/mozilla/nunjucks/pull/634).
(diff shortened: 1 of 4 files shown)


carljm:
Thanks! Do you have any reference for which browser version(s) support this behavior of `split()`? I wasn't able to find one on a quick search.

pbu88:
No I couldn't find any, it's so frustrating that it gives you that warning and let you clueless about which are the ones that don't support it, but my guess is that it should be supported by all modern browsers now a days because is a pretty standard way to use `split()` across different programming languages and 2nd, is described like that since the [ECMAScript 3rd edition (1999) reference](http://www.ecma-international.org/publications/files/ECMA-ST-ARCH/ECMA-262,%203rd%20edition,%20December%201999.pdf).

Do you maybe have a way to run the tests across multiple browsers?

carljm:
We don't have anything nice and automated set up for cross-browser testing, unfortunately, but I can check a few of the older supported browsers to verify.

pbu88:
Let me know how can I help you with that :)

carljm:
If it's part of ES3, I'm pretty sure it's fine. I'll test using Browserstack after I get this merged.

pbu88:
Cool! Thanks :)

PR #638, opened 2016-01-19
fix range template global to work with zero stop-value

devoidfury:
I'm using range to iterate up from a dynamic range, which is negative sometimes. This patch fixes the case that the stop-value is exactly zero. Added test case.

(This will also need a fresh browser build, not included)

### DIFF
--- a/src/globals.js
+++ b/src/globals.js
@@ -40,7 +40,7 @@ function joiner(sep) {
 function globals() {
     return {
         range: function(start, stop, step) {
-            if(!stop && stop !== 0) {
+            if(typeof stop === 'undefined') {
                 stop = start;
                 start = 0;
                 step = 1;
--- a/src/globals.js
+++ b/src/globals.js
@@ -40,7 +40,7 @@ function joiner(sep) {
 function globals() {
     return {
         range: function(start, stop, step) {
-            if(!stop) {
+            if(!stop && stop !== 0) {
                 stop = start;
                 start = 0;
                 step = 1;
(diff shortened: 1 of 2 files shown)


src/globals.js
```diff
@@ -40,7 +40,7 @@ function joiner(sep) {
 function globals() {
     return {
         range: function(start, stop, step) {
-            if(!stop) {
+            if(!stop && stop !== 0) {
```
carljm: Could we do `if(typeof stop === 'undefined')` here instead? Seems to communicate the actual intent more clearly.

carljm:
Thanks! Yes, new browser builds always happen at release time.

src/globals.js
devoidfury: What about the `null` / `false` cases?

carljm: Range isn't documented as accepting `null` or `false` for these arguments, and there's no reason to pass them. The purpose of this clause is just to handle the case where only one or two arguments are provided, and in that case the not-provided arguments are `undefined`.

devoidfury: Sure thing, I'll add the requested change then. My main concern was breaking backwards compatibility as little as possible.

carljm: Yep, I appreciate the motivation to preserve maximum back-compat. But back-compat for undocumented edge-cases always has to be weighed against the likelihood of someone actually relying on the current behavior (for reasonable reasons), and I just don't see it here.

devoidfury:
Thanks for the quick turnaround, I appreciate your work here!

PR #641, opened 2016-01-20
Support named `endblock` tags, fix #521

### LINKED ISSUE #521: Add option to name enblocks
It would be very useful can give a name for endblocks for help to identify them more easily like in twig

ricordisamoa:
As in Jinja2:
http://jinja.pocoo.org/docs/dev/templates/#named-block-end-tags

Also added two tests


carljm:
Looks great, thanks! Merged.

PR #644, opened 2016-01-21
environemnts dev option is not working

avcs06:
environemnt's dev option is not working, because it is being assigned to env.opts.dev but trying to read it from env.dev in line number 436, 468 and 485

### DIFF
--- a/src/environment.js
+++ b/src/environment.js
@@ -31,7 +31,7 @@ var Environment = Obj.extend({
         //  the library)
         // defaults to false
         opts = this.opts = opts || {};
-        this.opts.dev = !!opts.dev;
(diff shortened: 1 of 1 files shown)


carljm:
Hi! Thanks for the report and pull request. I think I'd like to keep the options all in `env.opts`, for consistency with the rest of the options. So I'd prefer to fix this the other way around -- fix the other references to use `env.opts.dev`.

carljm:
Fixed in 0b2fb96ae27e9ad34d32e7b1786cbc6ece190603 and 9c8b83806b99e53951214d726b9cff6035c25871 -- thanks for the report and draft patch!

PR #645, opened 2016-01-21
call configure() when the module initially loads

paulovieira:
The user will likely call `Nunjucks.render()` (or `renderString`) hundreds of times per second. There's no need to always be checking if we have an environment. We will be needing one anyway, so we might as well create a default environment from the beginning. 

Of course, the user can still override the default one by calling `Nunjucks.configure()` again.

### DIFF
--- a/index.js
+++ b/index.js
@@ -57,26 +57,17 @@ module.exports.configure = function(templatesPath, opts) {
     return e;
 };
 
+module.exports.configure();
+
 module.exports.compile = function(src, env, path, eagerCompile) {
-    if(!e) {
-        module.exports.configure();
-    }
     return new module.exports.Template(src, env, path, eagerCompile);
 };
 
 module.exports.render = function(name, ctx, cb) {
-    if(!e) {
-        module.exports.configure();
-    }
-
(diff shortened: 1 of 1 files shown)


ricordisamoa:
https://github.com/mozilla/nunjucks/issues/531#issuecomment-143820200
> In the longer run, I think the best/simplest option might be just to deprecate nunjucks.render entirely, and require always creating an environment and calling env.render.

carljm:
Yeah, I have mixed feelings about this. I think the "right" way to use nunjucks is to call `configure` yourself and use the resulting environment object, and avoid the top-level utility functions. So I'm a bit hesitant to penalize that usage (with an extra unnecessary environment created and thrown away on every page load) in order to facilitate the simplistic use of the top-level functions.

Basically: if you care enough about performance to need this PR, then you know enough to just create your own environment instead.

Thanks for the suggestion and pull request, but I'm going to close it.

paulovieira:
It might be 0.000001% more performant with this pr. But the value is in making the code in `index.js` more clear, in my opinion. Thanks for reviewing anyway.

PR #646, opened 2016-01-21
Tests may fail on Windows due to line ending issues

jrehwaldt:
This commit enforce LF line endings on checkout for files in `tests/**` in order to guarantee test success on Windows regardless of the user's git settings.


carljm:
Thanks!

Issue #561, opened 2015-10-24
Inconsistent scoping behavior in "macro", "include" and template inheritance.


legutierr:
This template...

    {% macro setval() %}{% set val = 22 %}{% endmacro %}{{ setval() }}{{ val }}

...evaluates to "" (i.e., the empty string).  The value set inside the macro doesn't leak out of the the local macro scope.  However, this template...

    {% macro setval() %}{% set val = 22 %}{% endmacro %}{% set val = 44 %}{{ setval() }}{{ val }}

...evaluates to "22". The value set inside the macro leaks out of the macro scope.  In Jinja, these two templates evaluate to "" and "44", respectively.

legutierr:
There's a similar weirdness happening with the include tag as well.  Given this simple template...

    {# template.html #}
    {% set var = 22 %}

...you get two very different results when you include it.  First...

    {% include "template.html" %}{{ var }}

...evaluates to "" (the empty string).  The value of var does not leak out of the included template's scope.  This template, however, does leak its scope...

    {% set var = 44 %}{% include "template.html" %}{{ var }}

...evaluating to "22".  In Jinja, these two templates evaluate to "" and "44", respectively, as before.

carljm:
Thanks for tracking this down! This is definitely something that should ideally be fixed, though without diving into the details I have no idea how invasive the fix will be. It's something I'd enjoy looking into if I find some time, but not sure when that will be - pull requests definitely welcome! (Even just a PR with failing tests for these issues would be a really good start.)

legutierr:
I discovered more of the same weirdness with template inheritance.  Given this base template...

    {# base.html #}
    {% block k %}{% set var = 'parent' %}{% endblock %}

...when you execute the following template...

    {% extends “base.html” %}{% block k %}{% set var = 'child' %}{{ super() }}{{ var }}{% endblock %}

...you get the expected result: ‘child’.  This is the same value that Jinja emits.  However, when you execute the following template...

    {% extends “base.html” %}{% block k %}{{ super() }}{{ var }}{% endblock %}

...you get ‘parent’.  Again, the problem is a leaking scope (in this case, that of “super()”), but only when the value is set.  With Jinja, in this second case, you get ‘’ (i.e. the empty string).

And there is also a fourth inconsistency.  Given this base template...

    {# base2.html #}
    {% set val = 'parent' %}{% block k %}{% set val = ‘inner’ %}{% endblock %}{{ val }}

...when you execute the following template...

[...]

PR #653, opened 2016-01-28
Do not resolve up when setting a variable in macro.

### LINKED ISSUE #577: Macro recursion issue: Variables in macros are not stored on the stack
Consider the following template that renders a list of fully qualified property names from a JSON schema:

    {% macro typeTable(node, prefix = "", name = "" %}
      {% set nodeName = prefix + ("." if prefix) + name) %}
      {% if nodeName %}
        <tr>
          <td><code>{{ nodeName | safe }}</code></td>
          <td>{{ node.type }}</td>
          <td>
            {% markdown %}{{ node.description }}{% endmarkdown %}
          </td>
        </tr>
      {% endif %}
      {% if node.type == "object" %}
        {% for propName, propSpec in node.properties %}
          {{ typeTable(propSpec, nodeName, propName) }}
        {% endfor %}
      {% endif %}
    {% endmacro %}

    <body>
      <table class='table table-bordered'>
        <thead>
          <tr>
            <th>Qualified name</th>
            <th>Type</th>
            <th>Description</th>
          </tr>
        </thead>
        <tbody>
          {{ typeTable(schema) }}
        </tbody>
      </table>
    </body>

Where `schema` is for example the following JSON schema

    {
      "$schema": "http://json-schema.org/draft-04/schema",
      "type": "object",
      "properties": {
        "id": {
          "type": "string",
          "description": "The ID of the node"
        },
        "name": {
          "type": "string",
          "description": "The name of the node"
        },
        "status": {
          "type": "object",
          "description": "The status of the node",
          "properties": {
            "vm": {
              "type": "string"
            },
            "app": {
              "type": "string"
            }
          }
        },
        "url": {
          "type": "string",
          "format": "uri",
          "description": "The URL of the node"
        }
      }
    }

The expected result is a table with the following "qualified names":

    id
    name
    status
    status.vm
    status.app
    url

However the variable `nodeName` seems to scoped per declaration (i.e. there is only one instance) and not scoped per invocation. This means that changes to `nodeName` in recursive calls of `typeTable` can be seen by the caller. This results in the following "qualified names":

    id
    id.name
    id.name.status
    id.name.status.vm
    id.name.status.vm.app
    id.name.status.vm.app.url

oyyd:
This PR should fix #577 and the "macro" part of issue #561.

### DIFF
--- a/src/runtime.js
+++ b/src/runtime.js
@@ -7,11 +7,11 @@ var Obj = require('./object');
 // we know how to access variables. Block tags can introduce special
 // variables, for example.
 var Frame = Obj.extend({
-    init: function(parent, inMacro) {
+    init: function(parent, isolate) {
         this.variables = {};
         this.parent = parent;
         this.topLevel = false;
-        this.inMacro = inMacro;
+        this.isolate = isolate;
     },
 
     set: function(name, val, resolveUp) {
@@ -21,7 +21,7 @@ var Frame = Obj.extend({
         var obj = this.variables;
         var frame = this;
 
-        if(resolveUp && !frame.inMacro) {
+        if(resolveUp && !frame.isolate) {
             if((frame = this.resolve(parts[0]))) {
                 frame.set(name, val);
                 return;
@@ -67,8 +67,8 @@ var Frame = Obj.extend({
         return p && p.resolve(name);
     },
 
-    push: function(inMacro) {
-        return new Frame(this, inMacro);
+    push: function(isolate) {
+        return new Frame(this, isolate);
     },
 
     pop: function() {
--- a/src/runtime.js
+++ b/src/runtime.js
@@ -7,10 +7,11 @@ var Obj = require('./object');
 // we know how to access variables. Block tags can introduce special
 // variables, for example.
 var Frame = Obj.extend({
-    init: function(parent) {
+    init: function(parent, inMacro) {
         this.variables = {};
         this.parent = parent;
         this.topLevel = false;
+        this.inMacro = inMacro;
     },
 
     set: function(name, val, resolveUp) {
@@ -20,7 +21,7 @@ var Frame = Obj.extend({
         var obj = this.variables;
         var frame = this;
 
-        if(resolveUp) {
+        if(resolveUp && !frame.inMacro) {
             if((frame = this.resolve(parts[0]))) {
                 frame.set(name, val);
                 return;
@@ -66,8 +67,8 @@ var Frame = Obj.extend({
         return p && p.resolve(name);
     },
 
-    push: function() {
-        return new Frame(this);
+    push: function(inMacro) {
+        return new Frame(this, inMacro);
     },
 
     pop: function() {
--- a/tests/compiler.js
+++ b/tests/compiler.js
@@ -485,6 +485,27 @@
                   { name: 'james' },
                   'FooInclude james');
 
+            equal('{% macro foo() %}{% set x = "foo"%}{{ x }}{% endmacro %}' +
+                  '{% set x = "bar" %}' +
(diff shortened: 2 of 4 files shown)


src/runtime.js
```diff
@@ -7,10 +7,11 @@ var Obj = require('./object');
 // we know how to access variables. Block tags can introduce special
 // variables, for example.
 var Frame = Obj.extend({
-    init: function(parent) {
+    init: function(parent, inMacro) {
```
carljm: What do you think about using a more generic name for this new parameter; something like `isolate`? It seems likely to me that this same parameter might prove useful in solving the other portions of #561, in which case the name `inMacro` might no longer apply.

tests/compiler.js
```diff
@@ -485,6 +485,27 @@
...
 
+            equal('{% macro foo() %}{% set x = "foo"%}{{ x }}{% endmacro %}' +
+                  '{% set x = "bar" %}' +
+                  '{{ x }}' +
+                  '{{ foo() }}' +
+                  '{{ x }}',
+                  'barfoobar');
+
+            // this test checks that the issue #577 is resolved
+            equal('{% macro foo(topLevel = "false", prefix = "") %}' +
```
carljm: Using the string "false" as the default parameter here seems needlessly misleading (since the string "false" is actually truthy). Either eliminate the default (since you aren't actually relying on it anywhere, or make it just `false` instead of `"false"`. Also, no spaces around the `=` in parameter lists.

tests/compiler.js
```diff
@@ -485,6 +485,27 @@
...
+                  '{% if topLevel %}' +
+                  '{% for i in [1,2,3] %}' +
+                  '{{ foo(false, x) }}' +
+                  '{% endfor %}' +
+                  '{% else %}' +
+                  '{{ x }}' +
+                  '{% endif %}' +
+                  '{% endmacro %}' +
+                  '{{ foo(true) }}',
+                  'foofoofoofoofoofoo');
```
carljm: I think this test could really use a brief comment explaining what the failure mode would look like. Just something along these lines: 

> If the bug is not fixed, and set variables leak into the caller scope, there will be too many "foo"s here, because each recursive call will append a "foo" to the variable `x` in its caller's scope, instead of just its own.

carljm:
This is great!! Thank you so much for all your work on nunjucks.

I made a few small comments above. Also, while you're updating the PR anyway, maybe add an entry to changelog? Thanks!

carljm:
(The changelog comment can just be something like "Fix `{% set %}` scoping within macros. Fixes #577 and the macro portion of #561. Thanks Ouyang Yadong. Merge of #653." -- where all the issue and PR links are Markdown links.)

carljm:
(Also, please comment here if you push another commit, GitHub doesn't notify when new commits are pushed to a PR.)

src/runtime.js
carljm: Though since it's internal, we can also leave it as-is for now and rename it later if necessary; I'm fine with that if you prefer.

oyyd:
Hi, @carljm. Thanks for your detailed comments which help a lot! :-) And I have add another test to make the test case more clear.

Maybe you can add me as a collaborator and I would like to help to improve Chinese docs(merge some PRs or make some PRs by myself). Simply ignore it if you don't think that's appropriate .

carljm:
@oyyd Sure, I'd be happy to add you as a collaborator to work on Chinese docs. I'd prefer if you get a review from someone before merging code that touches `src/*` (I do the same myself).

carljm:
Looks great, thank you!

oyyd:
@carljm Thank you. And I will continue to do PRs in other situations.

Issue #582, opened 2015-11-10
Block set


ElChupacabra26:
Obviously this would be an enhancement, but a block setter like they have in [Jinja](https://github.com/mitsuhiko/jinja2/issues/232) would be super useful. For instance, we could do things like this:

```jinja
{# Set modal content #}
{% set modalContent %}
    {% include 'partials/sku-page/bundle.nunj' %}
{% endset %}

{# Cache HTML for 'More Choices' modal in hidden field #}
<input type="hidden" class="modalContent" value="{{modalContent | e}}">
```

carljm:
Yes, I think nunjucks should maintain Jinja compatibility here. PR gladly accepted :-)

carljm:
Closed with merge of #656 -- thanks @daniele-rapagnani !

ElChupacabra26:
:+1: to that!

jtlapp:
My apologies, but I need this feature to use nunjucks at all, but it's not in the latest NPM release, and I don't see how to build the source. Is this in the docs somewhere? Thank you! (I have child templates that create variables from variables for use in multiple places in both the parent and child templates.)

carljm:
Yes, instructions for building are in the MAINTENANCE doc (really its as simple as `npm run browserfiles`). I'm hoping to get a new release out soon.

jtlapp:
Okay thank you. I guess I was thrown by the term "browserfiles" and the language "ready-made files for the browser." But of course if it works in the browser, it'll work on the server. In any case, I found the necessary soup of curly brackets confounding the readability of the variable assignments and am now opting for front matter JSON and two-pass variable assignment. Thanks for your help.

PR #656, opened 2016-01-30
Adds set block assignments a la jinja2

daniele-rapagnani:
I've implemented set block assignments as requested in #582.
I've updated the docs accordingly, taking inspiration by the jinja documentation.


browser/nunjucks-slim.js
```diff
@@ -1,4 +1,4 @@
-/*! Browser bundle of nunjucks 2.3.0 (slim, only works with precompiled templates) */
+/*! Browser bundle of nunjucks 3.0.0-dev.1 (slim, only works with precompiled templates) */
```
carljm: Please revert the changes to the browserfiles; that's done only at release time, otherwise it causes too many conflicts.

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@ Changelog
 
 master (unreleased)
 -------------------
+* Add support for `{% set %}` block assignments as in jinja2.
```
carljm: Can you add a link to the PR here, like the other changelist entries? Thanks! (You can also add a thanks to yourself, if that doesn't feel too weird! Think of it as the project thanking you, you're just doing the data entry :-) )

Also, blank line between header and first list element.

carljm:
Very nicely done, thanks so much for the PR! A couple minor comments above, ping me when they're addressed and I'll press the button!

daniele-rapagnani:
Ok, that should do it! 
Thanks for the corrections, let me know if I missed anything else.

PR #667, opened 2016-02-04
Fix scoping issues with macros and includes.

### LINKED ISSUE #666: Macros can see variables from their caller's scope.
In nunjucks currently, macros can see variables from their caller's scope, that were not explicitly passed in as arguments. E.g. the following template:

```
{% macro one(var) %}
  {{ two() }}
{% endmacro %}

{% macro two() %}
  {{ var }}
{% endmacro %}

{{ two("hello") }}
```

renders `hello`; macro `two` is able to see the value of `var` from macro `one`, even though it is not passed in as an argument.

 In Jinja2 that template renders nothing.

I consider the Jinja2 behavior correct, and the nunjucks behavior a bug.

It is possible, however, that people have written templates that depend on this bug to render correctly. For that reason, I'll probably fix it in 3.0 rather than in the 2.x series.

carljm:
This improves the fix for #577 and #561, expanding the fix for the latter to also cover includes, as well as macros.

The previous fix for #577 (PR #653) didn't take into account nested scopes within macros, which would happily bypass the `isolate` property of their parent frame and go right up into the calling frame.

A better fix for that is to make the macro frame actually completely isolated from the calling frame, breaking the parent relationship. This also fixes #666.

For includes (#561) we want included templates to _see_ variables in the including scope, but we don't ever want to write to the including scope. So we still need an `isolateWrites` attr on the frame, but it's specific to writes only, and (unlike in #653) it is respected not only at the originating level of the write, but also if a write from a child frame tries to go through it to its parent.


carljm:
@oyyd I'd welcome your review on this, if you have time.

src/compiler.js
```diff
@@ -831,7 +831,8 @@ var Compiler = Object.extend({
             '[' + argNames.join(', ') + '], ',
             '[' + kwargNames.join(', ') + '], ',
             'function (' + realNames.join(', ') + ') {',
-            'frame = frame.push(true);',
+            'callerFrame = frame;',
```
oyyd: Should this be `var callerFrame = frame;`? Global variables may cause potential bugs.

oyyd:
Hi @carljm. Everything looks good except the line of my comment. Good job :+1:

src/compiler.js
carljm: Yes, good catch! That was just an oversight on my part. Will fix.

carljm:
Thanks for the review; corrected the accidental global variable. Merging.

PR #683, opened 2016-02-19
fixed bad character leading % in {% endraw %}

dkebler:
just a bad version of the % in that line of docs, must have been some weird special character version.
For those copy and pasting the example the block would not close.


carljm:
Thanks for the fix!

Issue #678, opened 2016-02-17
Errors in included templates are thrown asynchronously


nicolasartman:
As far as I can tell, this applies to any block error and possibly any error at all

I've attached a folder ([nunjucks-repro.zip](https://github.com/mozilla/nunjucks/files/133691/nunjucks-repro.zip)) with the reproduction for convenience, and here's the code in each file:

**repro.js**

```
var nunjucks = require('nunjucks');

try {
    console.log('compiling');
    var template = nunjucks.compile('{% include "included.html" %}', new nunjucks.Environment([
        new nunjucks.FileSystemLoader('./')
    ]));
    console.log('rendering');
    template.render();
    console.log('done rendering');
} catch (err) {
    console.log('caught nunjucks error', err.message);
}
```

**included.html**

```
{% invalidTag %}
```

I wasn't able to figure out exactly why this is happening, but it appears the flow goes into an asyncIter (despite nothing async being used anywhere) [here](https://github.com/mozilla/nunjucks/blob/master/src/environment.js#L242), and then when it calls render again, it has passed a callback, causing [this](https://github.com/mozilla/nunjucks/blob/master/src/environment.js#L469) path to be taken, which erroneously reports the error asynchronously.

I hacked it in my fork by just throwing instead of ever calling callbackAsap, but that's clearly not the right way to fix it.

thanks!

carljm:
Any chance you could rework that example into a failing test in the nunjucks test suite? That'd be the next step towards fixing this.

nicolasartman:
Done!  In trying to make a failing test, I also discovered that it doesn't happen with invalid macros, but I'm not sure why. I thought that calling Template.render with a callback was the issue, but I see that's done anytime there's an import or include, and in most cases it works fine (even when there's an error such as the invalid macro) because compile doesn't throw during one of the calls where there's a callback, so it doesn't get reported asynchronously.

You can see the test [here](https://github.com/nicolasartman/nunjucks/commit/ee03e7900849f3abb5c4616f57926ec260afeada)—let me know if you want me to issue a PR or anything else.

carljm:
Yeah, go ahead and make a PR with that failing test, and mention this ticket number in its description. That makes it more visible / easier to find, the next time I get a chance to spend some time working on nunjucks. (Obviously if you're able to figure out a fix in the meantime and add it to the PR, even better!)

Thanks!

oleics:
Half an hour of debugging an express-app brought up two lines:

[...]

PR #689, opened 2016-02-22
Add test for async error throwing when includes have invalid tags

nicolasartman:
A failing test to demonstrate #678.

If there's a better place to put this test or preferred way to structure it just let me know. I just tried to mimic what's there already.


carljm:
Test looks great, thanks!

Issue #475, opened 2015-07-03
Extensionless extends, includes and imports


ArmorDarks:
Is there any way to provide nunjucks with templates files extension name, so we could drop it in extends, includes and imports.

For example, Instead of

``` twig
{% extends "layout.nj" %}
```

use

``` twig
{% extends "layout" %}
```

I've heard that it's possible to configure extension name when Nunjucks used in conjuration with Express. But what about other use cases, when Nunjucks used on it's own, or with Grunt?

jlongster:
Expresses imposes a weird relationship with a templating engine, as it wants to do these kinds of things itself. I'm not against supporting it natively though. If you would like to add it, open a PR.

carljm:
Closing this issue due to lack of follow-up. I think it's low priority, and it doesn't exist in Jinja2.

ArmorDarks:
:( I hopped that there is easy way to hook up it from Express-only to regular Nunjucks, but, unfortunately, didn't have enough time to look into it.

ArmorDarks:
@carljm posted in https://github.com/mozilla/nunjucks/pull/691#issuecomment-215252873:

> I still don't like #475 -- I think the recommended extension should remain a matter only for the documentation and the community, not something that has an actual effect on the behavior. Even if it were configurable, I still don't see the benefit -- it makes your includes etc correspond less clearly to the actual template files, why is that a good thing?

It's very minor thing, I agree. But it's just somehow wrong to ask developer specify extension every single time while importing or including something when you're clearly working in unified environment.

I think closest example here is ES2016, babel, systemjs. When you're importing other modules, you do not specify `.js` extension, unless you really need to specify extension for some serious reason. For example, when importing non-standard for system file. In other words, not `.js` one, but Sass or CSS. or image. It's just a good design, since system arguably assumes that you by default would require specific to current environment files only, and not anything else. In some sense, it DRYing out code, by removing constantly repeating and really not useful extension.

Same principle used in many other modern systems.

Option to drop extension just makes Nunjucks more unified with them.

carljm:
I think nunjucks already provides the flexibility necessary to implement this, though; it would just require a custom loader.

ArmorDarks:
Just to ensure — you don't see it as part of Nunjucks core? Otherwise, would be better to reopen issue.

carljm:
[...]

PR #691, opened 2016-03-03
add text for adoption of .njk extension

dkebler:
with this small edit it was easier to just start over with your latest commit with multiple edits. of templating.md.  Also made all textual lowercase nunjucks to Nunjucks

### DIFF
--- a/docs/templating.md
+++ b/docs/templating.md
@@ -6,13 +6,21 @@ title: Templates
 
 # Templating
 
-This is an overview of the templating features available in nunjucks.
+This is an overview of the templating features available in Nunjucks.
 
 > Nunjucks is essentially a port of
 > [jinja2](http://jinja.pocoo.org/docs/), so you can read their
 > [docs](http://jinja.pocoo.org/docs/templates/) if you find anything
 > lacking here. Read about the differences
 > [here](http://mozilla.github.io/nunjucks/faq.html#can-i-use-the-same-templates-between-nunjucks-and-jinja2-what-are-the-differences).
+
+## File Extensions
+
+Although you are free to use any file extension you wish for your
+Nunjucks template, files the Nunjucks community has adopted  `.njk`.  
+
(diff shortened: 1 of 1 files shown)


carljm:
Thanks!

ArmorDarks:
Great, though, we always used just `*.nj` for years...

Maybe now would be appropriate time to overlook https://github.com/mozilla/nunjucks/issues/475?

carljm:
Nothing wrong with continuing to use `.nj` if you prefer it, but I think it's useful to have a documented preference. No matter what we picked, it wouldn't match what someone out there has been using...

I still don't like #475 -- I think the recommended extension should remain a matter only for the documentation and the community, not something that has an actual effect on the behavior. Even if it were configurable, I still don't see the benefit -- it makes your includes etc correspond less clearly to the actual template files, why is that a good thing?

ArmorDarks:
> Nothing wrong with continuing to use .nj if you prefer it, but I think it's useful to have a documented preference. No matter what we picked, it wouldn't match what someone out there has been using...

Sure. Just wonder, why was chosen `.njk`? I didn't find any discussion about it. Did I miss something? Just curios.

Anyways, it definitely not that important, but just my two cents. It seems strange to have ending `k`. It's like to have `.jst` for JavaScript just because it ends with `t`.

> I still don't like #475 -- I think the recommended extension should remain a matter only for the documentation and the community, not something that has an actual effect on the behavior. Even if it were configurable, I still don't see the benefit -- it makes your includes etc correspond less clearly to the actual template files, why is that a good thing?

I've replied here https://github.com/mozilla/nunjucks/issues/475#issuecomment-215259279 to avoid off-topic.

carljm:
It was discussed briefly on the Google group mailing list. I said I didn't care between `nj` and `njk`, and would leave it up to whoever submitted a PR to paint the bikeshed. I have a slight preference toward `nj` myself, but I don't really care.

PR #701, opened 2016-03-17
Fix to issue #700; should eliminate double-escaping problem.

### LINKED ISSUE #700: Double escaping should be eliminated.
In Jinja, you can apply an escape filter on any string data multiple times, and it will not change the output, regardless as to whether auto escape is turned on or not:

    >>> env = jinja2.Environment(autoescape=False);
    >>> env.from_string("{% set val = '<a>' %}{% set var = val|e %}{{ var }}").render({})
    u'&lt;a&gt;'
    >>> env.from_string("{% set val = '<a>' %}{% set var = val|e %}{{ var|e }}").render({})
    u'&lt;a&gt;'
    >>> env = jinja2.Environment(autoescape=True);
    >>> env.from_string("{% set var = '<a>' %}{% set val = var %}{{ val }}").render({})
    u'&lt;a&gt;'
    >>> env.from_string("{% set var = '<a>' %}{% set val = var|e %}{{ val }}").render({})
    u'&lt;a&gt;'
    >>> env.from_string("{% set var = '<a>' %}{% set val = var|e %}{{ val|e }}").render({})
    u'&lt;a&gt;'
    >>> env.from_string("{% set var = '<a>' %}{% set val = var|e %}{{ val|e|e|e }}").render({})
    u'&lt;a&gt;'

You get the idea.  Nunjucks, unfortunately, doesn't work this way:

    > var env = new nj.Environment(new nj.FileSystemLoader(''), {'autoescape':false});
    undefined
    > env.renderString("{% set val = '<a>' %}{% set var = val|e %}{{ var }}");
    '&lt;a&gt;'
    > env.renderString("{% set val = '<a>' %}{% set var = val|e %}{{ var|e }}");
    '&amp;lt;a&amp;gt;'
    > var env = new nj.Environment(new nj.FileSystemLoader(''));
    undefined
    > env.renderString("{% set val = '<a>' %}{% set var = val|e %}{{ var }}");
    '&amp;lt;a&amp;gt;'
    > env.renderString("{% set val = '<a>' %}{% set var = val|e %}{{ var|e }}");
    '&amp;amp;lt;a&amp;amp;gt;'

The problem seems to be that escaped strings are not immediately marked as safe after they are generated, but before they are returned.  That would seem fix the problem:

    > env.renderString("{% set val = '<a>' %}{% set var = val|e|safe %}{{ var|e|e|e|e }}");
    '&lt;a&gt;'

A pull request will be forthcoming.

legutierr:
Should fix #700

### DIFF
--- a/src/filters.js
+++ b/src/filters.js
@@ -117,7 +117,7 @@ var filters = {
 
     escape: function(str) {
         if(typeof str === 'string') {
-            return lib.escape(str);
+            return r.markSafe(lib.escape(str));
         }
         return str;
     },
--- a/tests/filters.js
+++ b/tests/filters.js
@@ -98,6 +98,33 @@
             finish(done);
         });
 
+        it('should not double escape strings', function(done) {
+            var res = render('{{ "<html>" | escape | escape }}', {}, { autoescape: false });
+            expect(res).to.be('&lt;html&gt;');
+            finish(done);
+        });
+
+        it('should not double escape with autoescape on', function(done) {
+            var res = render('{% set val = "<html>" | escape %}{{ val }}', {}, { autoescape: true });
+            expect(res).to.be('&lt;html&gt;');
+            finish(done);
+        });
+
+        it('should not escape safe strings with autoescape on', function(done) {
+            var res1 = render('{{ "<html>" | safe | escape }}', {}, { autoescape: true });
+            expect(res1).to.be('<html>');
+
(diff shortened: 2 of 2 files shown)


carljm:
Looks great, thanks! I'll merge tomorrow.

carljm:
Looks good to me; merging. Thanks!

PR #705, opened 2016-03-20
Support ECMAScript 2015 Map and Set in length filter

ricordisamoa:
They both expose a 'size' property:
http://www.ecma-international.org/ecma-262/6.0/#sec-get-map.prototype.size
http://www.ecma-international.org/ecma-262/6.0/#sec-get-set.prototype.size

Use that instead of the 'length' property.
Also added two tests that only run if the respective globals are set.

### DIFF
--- a/src/filters.js
+++ b/src/filters.js
@@ -175,7 +175,17 @@ var filters = {
     length: function(val) {
         var value = normalize(val, '');
 
-        return value !== undefined ? value.length : 0;
+        if(value !== undefined) {
+            if(
+                (typeof Map === 'function' && value instanceof Map) ||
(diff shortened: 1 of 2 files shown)


ricordisamoa:
It isn't very clear to me to what extent ES2015 is embraced by Nunjucks, but I made the PR anyway since they seemed roughly equivalent to Python dicts and sets.

carljm:
This looks good to me! I don't think nunjucks currently has much ES2015 support, but that's no reason not to start improving it. Thanks for the PR! I'm on vacation for the next two weeks, but I'll merge it when I get back

Issue #712, opened 2016-03-23
Nunjucks implementation of built-in `round` filter does not match jinja documentation


justinsthomas:
The Nunjucks docs for built-in filters (https://mozilla.github.io/nunjucks/templating.html#builtin-filters) point to the Jinja2 docs (`round`, for example: http://jinja.pocoo.org/docs/dev/templates/#round).

The issue I'm encountering is as follows:

`{{ 3.4545 | round }}` returns `3.0` with jinja
`{{ 3.4545 | round }}` returns `3` with nunjucks

According to the jinja docs, I should have to append the `int` filter (`{{ 3.4545 | round | int }}`) to return `3`, but this is not the case. This should be able to be remedied by adding `.toFixed(1)` to the returned value of the `round` filter; pull request coming soon.

legutierr:
I don't think that it would be correct to call `.toFixed(1)` on the output of `| round` here.  The `round` filter should be outputting a number type, not a string type (`toFixed` outputs a string type).  You can see the problem with this in the following example: 

    {{ 1 + 2.5 | round }}

Currently, this expression outputs to '4', but if `.toFixed(1)` were called on the output of `round` before it returned, the expression would result in '13', because when one of the arguments to `+` is a string, the other argument is coerced to a string as well.

Unfortunately, this is more an issue of the way that number types print in JavaScript, and of the fact that JavaScript only has one number datatype, rather than Python's 2+ number datatypes (`int`, `float`, and then some).  Achieving correspondence between Jinja and Nunjucks on this issue from the Nunjucks side would probably require implementing stand-ins for Python's `int` and `float` datatypes inside Nunjucks, and carrying this differentiation across the entirety of the implementation.  It's a rabbit hole.

The way that I achieved correspondence in this kind of situation is by attacking it from the Jinja side.  Giving Nunjucks pseudo-ints and pseudo-floats so that it matches the behavior of Jinja is much harder than restricting Jinja to using `float` types in most cases, and printing them in a manner consistent with Javascript.  That way, in both Jinja and Nunjucks `{{ 3.4545 | round }}` will output `3`.  If you're interested in this code, we do plan on releasing it soon.

carljm:
[...]

PR #713, opened 2016-03-23
Give Nunjucks the ability to match properly output of Jinja's 'round' filter

justinsthomas:
* Added option 'jinjaCompat' param to 'round' filter to match jinja's default returned floating point
  i.e. `{{ 3.4545 | round(0, 'common', true) }}` will return `4.0` instead of `4`
* Added tests
* Updated docs to reflect changes

### DIFF
--- a/src/filters.js
+++ b/src/filters.js
@@ -317,8 +317,10 @@ var filters = {
         return arr;
     },
 
-    round: function(val, precision, method) {
+    round: function(val, precision, method, jinjaCompat) {
         precision = precision || 0;
+        jinjaCompat = jinjaCompat || false;
+
         var factor = Math.pow(10, precision);
         var rounder;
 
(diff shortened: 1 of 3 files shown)


src/filters.js
```diff
@@ -332,7 +334,14 @@ var filters = {
             rounder = Math.round;
         }
 
-        return rounder(val * factor) / factor;
+        var roundedValue = rounder(val * factor) / factor;
+
+        if(Number.isInteger(roundedValue) && jinjaCompat) {
```
ricordisamoa: `Number.isInteger()` is not implemented under Node.js 0.10

justinsthomas:
@ricordisamoa updated to be backwards-compatible with node 0.10

legutierr:
On cursory reading, this change looks like it will produce incorrect results if the results of the `round` filter are passed into another expression.  For instance, the following will result in '15.0', rather than '6':

    {{ 1 + 4.501 | round(0, "normal", true) }}

This is because the included change causes an output of string, rather than int.  Another inconsistency here is that it looks like you are outputting different data types depending on the number of rounding digits, so the following actually produces the correct output of '5.5':

    {{ 1 + 4.501 | round(1, "normal", true) }}

I've put some additional comments on this in the original ticket #712.

carljm:
Given the comments in #712, I'm not sure this is fixable at the level of the `round` filter.

Thanks @justinsthomas for the PR, and thanks @legutierr for the astute observations!

PR #716, opened 2016-03-30
endraw

t-a-r-o-o-o-o-o:
I fixed endraw error.

> nunjucks.min.js:2 Uncaught Template render error: (unknown path) [Line 5, Column 2]
  unknown block tag: endraw

### DIFF
--- a/src/parser.js
+++ b/src/parser.js
@@ -542,7 +542,9 @@ var Parser = Object.extend({
         }
 
         switch(tok.value) {
-        case 'raw': return this.parseRaw();
+        case 'raw': 
+        case 'endraw': 
+            return this.parseRaw();
         case 'if':
         case 'ifAsync':
             return this.parseIf();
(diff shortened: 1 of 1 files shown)


legutierr:
Hi @t-a-r-o-o-o-o-o.  

Do you have any tests to demonstrate what this pull request is fixing?  Also, I am unable to identify an issue describing the original problem.  Has it been documented in the issues, or elsewhere?

One note on the implementation: it looks like your line 456 is misplaced.  The `endraw` token should never be found at the beginning of a raw block, which is what that switch statement is looking for.  If parseStatement(), encounters an `endraw` tag, that *should* be an error, I think.  If I understand it properly, I think your code would permit blocks of the form `{% endraw %}{% endraw %}` without any `{% raw %}` tag at the beginning.

t-a-r-o-o-o-o-o:
Sorry, Line 546 is unnecessary.
It is my mistake.
I've fixed it.
Please check it.

legutierr:
Hi @t-a-r-o-o-o-o-o.  You are still missing tests :) 

It looks like the reason for you to make this change is because you are using custom block start and block end tokens.  Is that correct?  Maybe the way to test this is to show a specific situation where you are using nested `raw` blocks together with custom start and end block tokens.  

It actually doesn't look like there are any comprehensive tests of the custom tokens as described here in the documentation (with the exception of some basic tests of the lexer only):

http://mozilla.github.io/nunjucks/api.html#customizing-syntax

What do you think about contributing some tests that cover this feature, as well as the modification in your pull request?

uerbin:
unknown block tag: endraw 

not merge ?


front tempate && backend tempate 

i want change block tags

blockStart: '<%', blockEnd: '%>', 
variableStart: '<$', variableEnd: '$>', 
commentStart: '<#',commentEnd: '#>'


my gold , it does not work

uerbin:
when show  code 👍 

<% raw %>
	<h3><$ title $></h3>
<% endraw %>

Issue #734, opened 2016-04-22
include tag doesn't preserve indentation level on included text


nicolasartman:
Given the input
```
line one
  {% include "test.txt" %}
  another indented line
```
and test.txt contents: 
```
included line one
  included line two
```
I would expect everything included to have the same relative indentation, like so:
```
line one
  included line one
    included line two
  another indented line
```
but instead, only the first line is indented, and the rest are reset to 0 relative indent (so only the indentation amount present in the included file is left):
```
line one
  included line one
  included line two
  another indented line
```

This ends up causing a couple nasty interactions for me, such as using a markdown region (with the nunjucks-markdown tag) with a fenced code block—the stripped indentation doesn't display correctly inside the resultant `<pre><code>` bits. I think this will generally be an issue when trying to include code inside a <pre> with predictable indentation.

I created a failing test case which is in <pull request coming now>.

Thanks!

carljm:
Hmm. This is expected behavior (by me), but I can see how it could be unexpected in some contexts, especially if using includes with indentation-sensitive formats like Markdown.

I think I am slightly opposed to changing it, however, because a) the current behavior matches the behavior of Jinja2, and Jinja2 compatibility is generally high priority for me, and b) the current behavior is very simple: the results of rendering the included template are literally included as-is in the including template in place of the `include` tag, with no munging and no special treatment of newlines. Your proposed behavior would involve always treating newlines specially in an included template, and inserting new spaces after every newline. This might solve some use cases, but could easily break others (imagine an included template with some pre-formatted text in it).

If we did this, it would not be a change to the default behavior of `include`, but an optional argument to `include` (which would be false by default, maintaining the current behavior by default). I guess I would probably accept a PR for an implementation that takes that approach.

nicolasartman:
Gotcha, thanks! Do you have any pointers on where I'd look to try to implement this? (If you'd prefer I take that discussion to the mailing list or a chat somewhere I'm happy to).

carljm:
You'd probably want to start with `compileInclude` in `src/compiler.js`, though to add the optional argument you'll also need to modify `parseInclude` in `src/parser.js`, and probably modify the `Include` node type in `nodes.js` as well.

kriss145:
Hi,

I have similar problem. Below is example:

c.tpl
```
<ul class="list">
    <li class="item">
        <div class="feature">

        </div>
    </li>
</ul>
```

index.tpl
```
<div class="main">
    {% include "c.tpl" %}
[...]

PR #735, opened 2016-04-22
Add test for include tag not keeping indent on included text

nicolasartman:
A test to reproduce #734.


carljm:
As discussed on #734, I don't believe we should change the default behavior of `include` to match the behavior specified in this test. I would be open to a PR that implemented this behavior as an opt-in extra argument to `include`.

PR #746, opened 2016-05-05
Add support for strict type check comparisons

oughter:
I added support for the `===` and `!==` comparison operators.


carljm:
This is a departure from Jinja2 syntax, which normally I would prefer not to do, but in this case I buy that it's necessary; `==` already behaves differently in nunjucks vs Jinja2 because internally it's a Python `==` vs a JS `==`, and those are different. I don't believe this is a difference we will reasonably be able to paper over. And if you're limited to JS `==` semantics, then you need the option of `===` as well.

carljm:
If I were implementing nunjucks from scratch, I would be awfully tempted to have `==` in a nunjucks template always translate to `===` at the JS level, but I think that ship has sailed; it would be a seriously backwards-incompatible change.

ricordisamoa:
Why not in a new major version?

carljm:
@ricordisamoa I'm open to that, if you want to open a ticket for it and put it in the 3.0 milestone.

Hmm... I guess if we plan to do that, it might be better not to merge this? It would be silly to end up in a situation where we support both `==` and `===`, but both actually mean `===`.

On the other hand, if it's in 3.0 we can always just drop `===` at that point again...

oughter:
@carljm, thanks for going over my changes. I understand the Jinja2 influence and desire not to stray from that. I hope that a benefit from this new functionality is a lower barrier to adoption for swig users looking to migrate their templates since swig is no longer maintained. Nunjucks seems like the natural choice for migrating swig templates, and swig has strict type checking.

I’m in the process of migrating some swig templates to nunjucks and ran into `===` not compiling for nunjucks. It’s not a big deal to replace `===` with `==`, but some time needs to be spent making sure that the comparison doesn’t actually need to be type strict, or if `===` was used because developers have been beaten over the head by linters to use that whenever they write JS (and writing templates can feel like writing regular JS).

ricordisamoa:
> I'm open to that, if you want to open a ticket for it and put it in the 3.0 milestone.

#750

ArmorDarks:
There is a typo in release notes:

> Add support for strict type check comparisons (=== and !===). Thanks

`!===` should be `!==`

I also hope that #750 is still valid for 3.0.0 version?

Issue #759, opened 2016-05-13
add for in support for es6 maps and sets


nbransby:
as length filter has been fixed to work with these classes I'm guessing this is on the cards soon?

carljm:
It'd be a welcome improvement. How soon it's in the cards really depends how soon someone implements it and submits a pull request; keeping up with issues and pull requests is about all I have time for.

ricordisamoa:
Yes it should be done

ricordisamoa:
@fdintino @maxnordlund fixed by #1058?

maxnordlund:
Yes I believe so.

PR #760, opened 2016-05-13
Support ECMAScript 2015 Set in for block #759

ricordisamoa:
Synchronous only.

### DIFF
--- a/src/compiler.js
+++ b/src/compiler.js
@@ -688,17 +688,40 @@ var Compiler = Object.extend({
             v = this.tmpid();
             frame.set(node.name.value, v);
 
-            this.emitLine('var ' + len + ' = ' + arr + '.length;');
-            this.emitLine('for(var ' + i + '=0; ' + i + ' < ' + arr + '.length; ' +
-                          i + '++) {');
-            this.emitLine('var ' + v + ' = ' + arr + '[' + i + '];');
-            this.emitLine('frame.set("' + node.name.value + '", ' + v + ');');
+            this.emitLine('if(runtime.isArray(' + arr + ')) {'); {
+                this.emitLine('var ' + len + ' = ' + arr + '.length;');
+                this.emitLine('for(var ' + i + '=0; ' + i + ' < ' + arr + '.length; ' +
+                              i + '++) {');
+                this.emitLine('var ' + v + ' = ' + arr + '[' + i + '];');
+                this.emitLine('frame.set("' + node.name.value + '", ' + v + ');');
 
-            this.emitLoopBindings(node, arr, i, len);
+                this.emitLoopBindings(node, arr, i, len);
 
-            this.withScopedSyntax(function() {
-                this.compile(node.body, frame);
-            });
+                this.withScopedSyntax(function() {
+                    this.compile(node.body, frame);
+                });
+
+                this.emitLine('}');
+            }
+
+            this.emitLine('} else if(runtime.isSet(' + arr + ')) {'); {
+                v = this.tmpid();
+                frame.set(node.name.value, v);
+
+                this.emitLine(i + ' = -1;');
+                this.emitLine('var ' + len + ' = ' + arr + '.size;');
+                this.emitLine('for(var ' + v + ' of ' + arr + ') {');
+                this.emitLine(i + '++;');
+                this.emitLine('frame.set("' + node.name.value + '", ' + v + ');');
+
+                this.emitLoopBindings(node, arr, i, len);
+
+                this.withScopedSyntax(function() {
+                    this.compile(node.body, frame);
+                });
+
+                this.emitLine('}');
+            }
 
             this.emitLine('}');
         }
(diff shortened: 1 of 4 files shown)


src/lib.js
```diff
@@ -106,6 +106,10 @@ exports.isArray = Array.isArray || function(obj) {
     return ObjProto.toString.call(obj) === '[object Array]';
 };
 
+exports.isSet = function(obj) {
+    return ObjProto.toString.call(obj) === '[object Set]';
```
ricordisamoa: What about polyfills? `[object Object]`...

src/compiler.js
```diff
@@ -688,17 +688,40 @@ var Compiler = Object.extend({
...
+                this.emitLine('}');
+            }
+
+            this.emitLine('} else if(runtime.isSet(' + arr + ')) {'); {
+                v = this.tmpid();
+                frame.set(node.name.value, v);
+
+                this.emitLine(i + ' = -1;');
+                this.emitLine('var ' + len + ' = ' + arr + '.size;');
+                this.emitLine('for(var ' + v + ' of ' + arr + ') {');
```
ricordisamoa: What if the environment does not have for...of syntax?

ricordisamoa:
I don't like it, it turned out unreadable, duplication and bad tests

src/compiler.js
ricordisamoa: it fails: https://travis-ci.org/mozilla/nunjucks/jobs/130114774

carljm: If we don't support polyfills, then I don't think we have to worry about that, do we? Any environment with real Sets will also have `for...of`, won't it?

src/compiler.js
```diff
@@ -688,17 +688,40 @@ var Compiler = Object.extend({
...
-            });
+                this.withScopedSyntax(function() {
+                    this.compile(node.body, frame);
+                });
+
+                this.emitLine('}');
+            }
+
+            this.emitLine('} else if(runtime.isSet(' + arr + ')) {'); {
+                v = this.tmpid();
```
carljm: do we need a different tmpid here? why not use the same one? the is-array and is-set clauses are mutually exclusive.

carljm:
I think you're being too hard on yourself :-) The for loop code was already hard to read (code-generation code in general is hard to read, and nunjucks generally is too). Your PR follows the same patterns as the existing code (don't worry about some duplication, prioritize speed over size) and makes sense to me. What's wrong with the tests?

carljm:
Looks like it's failing on Node 0.10, probably because it doesn't support `for...of`?

src/compiler.js
ricordisamoa: If the environment doesn't have real Sets, `runtime.isSet()` will always return `false` but the whole output of compilation must be parsed correctly even if it never runs.

carljm: oh, never mind, i'm wrong of course. because this syntax is generated as part of the compiled template regardless of whether a set is actually ever passed in. so yeah... I'm not sure how we can actually do this. Is there a way to iterate over sets without using `for...of`? Can we just convert to an array first?

ricordisamoa:
> probably because it doesn't support for...of

yes

src/compiler.js
ricordisamoa: cargo cult programming

src/compiler.js
ricordisamoa: `Set.prototype.keys()`, `Set.prototype.values()` and `Set.prototype.entries()` return `SetIterator` which apparently has a Python-like `.next()` method.

src/lib.js
shvaikalesh: What about `obj && obj[Symbol.toStringTag] == 'Set'`? Does not require shims to patch `Object#toString` (I would argue that they should, and most of them do), but requires `Symbol` polyfill.

src/compiler.js
shvaikalesh: `Set` is polyfillable, `for ... of` is not. I would argue we should avoid requiring Babel to run the code and provide a helper for iterating (boring) or rather require `Array.from` shim (it converts *iterables* to arrays and will simplify the branching).

maxnordlund:
Actually the code supports anything that happens to support the [iterator protocol](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols). But to be pragmatic, and support old browsers, why not use [Array.from](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from) directly.

The only caveat I can think of is providing a string, which would iterate over the individual characters. But then again, by then your in la la land anyways.

The code could be simplified as `arr = (typeof Array.from === "function") ? Array.from(arr) : arr`. Then just reuse the old flow. If you don't have time @ricordisamoa I can make a PR with that change, but I don't want to steal this PR from you.

ricordisamoa:
@maxnordlund steal it, please :grinning:

maxnordlund:
So Christmas happened, but I finally got around to doing this. Check it out and give me some feedback, pretty please?

maxnordlund:
This is done I think, ping @fdintino

fdintino:
It is. I'll wait until I merge in the develop branch and put out a release before I close out this issue though.

PR #761, opened 2016-05-16
Force unix style path separators in precompile

kim3er:
If you precompile in Windows, Nunjucks uses Windows path separators. This leads to inconsistent results, when precompiling between *NIX and Windows.

PR standardises path separators to *NIX style. All tests passing.

### DIFF
--- a/src/precompile.js
+++ b/src/precompile.js
@@ -88,6 +88,8 @@ function precompile(input, opts) {
         for(var i=0; i<templates.length; i++) {
             var name = templates[i].replace(path.join(input, '/'), '');
 
+            name = name.replace('\\', '/');
+
             try {
                 precompiled.push( _precompile(
                     fs.readFileSync(templates[i], 'utf-8'),
--- a/src/precompile.js
+++ b/src/precompile.js
@@ -88,7 +88,7 @@ function precompile(input, opts) {
         for(var i=0; i<templates.length; i++) {
             var name = templates[i].replace(path.join(input, '/'), '');
 
-            name = name.replace('\\', '/');
+            name = name.replace(/\\/g, '/');
 
             try {
                 precompiled.push( _precompile(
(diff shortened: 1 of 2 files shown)


src/precompile.js
```diff
@@ -88,6 +88,8 @@ function precompile(input, opts) {
         for(var i=0; i<templates.length; i++) {
             var name = templates[i].replace(path.join(input, '/'), '');
 
+            name = name.replace('\\', '/');
```
rhengles: This needs to be a regex, this only replaces the first slash.

carljm: It would be good to add a test that would have caught this.

kim3er:
Thanks for pointing that out, @rhengles. Sorted.

@carljm, I gave some thought to a test. Given that the issue is platform specific, is it acceptable that the test would only ever fail on Windows?

I was thinking of writing a test that ran a precompile, then checked the `wrapper` function for the correct pat separator, in the template argument.

carljm:
Yes, that's the sort of test that I was thinking of too. And yes, I think it's ok if it only fails on Windows. I guess we could write a test that would fail anywhere by mocking out some of the underlying path module? But that's harder, and I'm not convinced it's actually more useful. Better for the test to fail where the code would actually fail. We do have Windows CI.

Thanks!

rhengles:
IMHO, if we are standardizing on forward slashes, the test should simply disallow backwards slashes. If I understand this correctly.

kim3er:
@carljm: I've added a test that checks the output of `wrapper`.

@rhengles: The backwards slashes are symptomatic of the host operating system being Windows. The test checks the output of the `wrapper` function, which is fired, after my proposed fix has affected the change.

kim3er:
Is there anything else I can do to progress this one?

Lemmy4555:
This change caused issues to our projects since we used to write paths in windows style, we had to force an older version of nunjucks as a sub dependecy of gulp-nunjucks by using npm shrinkwrap and forcing version 2.4.2 of nunjucks.

`"nunjucks": {
          "version": "**2.4.2**",
          "from": "nunjucks@**2.4.2"**,
          "resolved": "https://registry.npmjs.org/nunjucks/-/nunjucks-**2.4.2**.tgz",
          "dependencies": {
           ...
           }
}`

Issue #764, opened 2016-05-16
Rename test files to `.njk` instead of `.j2`


niftylettuce:
Per https://mozilla.github.io/nunjucks/templating.html

> Although you are free to use any file extension you wish for your Nunjucks template files, the Nunjucks community has adopted .njk.
> 
> If you are developing tools or editor syntax helpers for Nunjucks, please include recognition of the .njk extension.

PR #780, opened 2016-06-10
Update file extensions to match .njk convention

oughter:
Per #764, renamed test files from ".j2" to ".njk". Also fixed a few spots from a previous find and replace from ".html" to ".j2".


carljm:
Thanks! This is actually a change that doesn't really need/deserve a changelog entry, since it has no impact on nunjucks users. Can you force-push your branch with just the actual renaming commit? Or if you're not sure how to do that in git, you could also just open a new PR. Thanks!

oughter:
@carljm, done!

carljm:
Thank you!

PR #789, opened 2016-07-10
Bump chokidar version to fix node 6.x deprecated call

tbouron:
Root cause comes from `fsevents`, [see PR](https://github.com/strongloop/fsevents/pull/128)

Before the bump
```
$ node --version
v6.3.0

$ node -e 'require("chokidar")'
(node) v8::ObjectTemplate::Set() with non-primitive values is deprecated
(node) and will stop working in the next major release.

==== JS stack trace =========================================

[... omitted]
=====================


==== C stack trace ===============================

 1: v8::Template::Set(v8::Local<v8::Name>, v8::Local<v8::Data>, v8::PropertyAttribute)
 2: fse::FSEvents::Initialize(v8::Local<v8::Object>)
 3: node::DLOpen(v8::FunctionCallbackInfo<v8::Value> const&)
 4: v8::internal::FunctionCallbackArguments::Call(void (*)(v8::FunctionCallbackInfo<v8::Value> const&))
 5: v8::internal::MaybeHandle<v8::internal::Object> v8::internal::(anonymous namespace)::HandleApiCallHelper<false>(v8::internal::Isolate*, v8::internal::(anonymous namespace)::BuiltinArguments<(v8::internal::BuiltinExtraArguments)1>)
 6: v8::internal::Builtin_HandleApiCall(int, v8::internal::Object**, v8::internal::Isolate*)
 7: 0x17987870961b
(node) v8::ObjectTemplate::Set() with non-primitive values is deprecated
(node) and will stop working in the next major release.

==== JS stack trace =========================================
[... omitted]
```

After, everything works fine, no more stacktrace, including for downstream projects

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "author": "James Long <longster@gmail.com>",
   "dependencies": {
     "asap": "^2.0.3",
-    "chokidar": "^1.0.0",
+    "chokidar": "^1.6.0",
     "yargs": "^3.32.0"
   },
   "browser": "./browser/nunjucks.js",


debugish:
This looks to already be on 3.x-dev (master) but would be great to be able to work with node 6.x on a stable tag, would it be appropriate for this land in a v2.4.3 as well?

BTW I tested and confirmed this fix with node 6.3.1 as well.

Would It be appropriate to create a separate issue for this concern?

carljm:
This fix is already on the 2.x branch and will be in 2.4.3 when it happens. I don't know when that'll be; when I have time to roll a release.

PR #798, opened 2016-07-22
Docs filter updates

vipickering:
I have added all the undocumented filters to the documentation file.

I've also tested each example to make sure it works.

### DIFF
--- a/docs/templating.md
+++ b/docs/templating.md
@@ -910,7 +910,7 @@ Return the absolute value of the argument.
 
 ### batch
 
-A filter that batches items.It returns a list of lists with the given number of items.
+Batch returns a list of lists with the given number of items:
 
 **Input**
 
@@ -931,7 +931,7 @@ A filter that batches items.It returns a list of lists with the given number of 
 
 ### capitalize
 
-Makes the first letter uppercase, the rest lower case.
+Makes the first letter uppercase, the rest lower case:
 
 **Input**
 
@@ -948,7 +948,7 @@ This is a test
 
 ### center
 
-Centers the value in a field of a given width.
+Centers the value in a field of a given width:
 
 **Input**
 
@@ -964,7 +964,7 @@ fooo
 
 ### dictsort
 
-Sort a dict and yield (key, value) pairs.
+Sort a dict and yield (key, value) pairs:
 
 ```jinja
 {% set items = {
@@ -989,8 +989,8 @@ a b c d e f
 ### escape (aliased as e)
 
 Convert the characters &, <, >, ‘, and ” in string s to HTML-safe sequences.
-Use this if you need to display text that might contain such characters in
-HTML. Marks return value as markup string.
+Use this if you need to display text that might contain such characters in HTML.
+Marks return value as markup string
 
 **Input**
 
@@ -1006,8 +1006,8 @@ HTML. Marks return value as markup string.
 
 ### float
 
-Convert the value into a floating point number. If the conversion doesn’t work
-it will return 0.0. You can override this default using the first parameter.
+Convert a value into a floating point number. If the conversion fails 0.0 is returned.
+This default can be overridden by  using the first parameter.
 
 **Input**
 
@@ -1024,7 +1024,7 @@ it will return 0.0. You can override this default using the first parameter.
 
 ### first
 
-Get the first item in an array.
+Get the first item in an array:
 
 **Input**
 
@@ -1042,7 +1042,7 @@ Get the first item in an array.
 
 ### groupby
 
-Group a sequence of objects by a common attribute.
+Group a sequence of objects by a common attribute:
 
 **Input**
 
@@ -1072,9 +1072,9 @@ blue : john jim
 
 ### indent
 
-Return a copy of the passed string, each line indented by 4 spaces. The first
-line is not indented. If you want to change the number of spaces or indent
-the first line too you can pass additional parameters to the filter:
+Indents a string using spaces.
+Default behaviour is *not* to indent the first line.
+Default indentation is 4 spaces.
 
 **Input**
 
@@ -1088,6 +1088,38 @@ the first line too you can pass additional parameters to the filter:
 one
     two
     three
+```
+
+Change default indentation to 6 spaces:
+
+**Input**
+
+```jinja
+{{ "one\ntwo\nthree" | indent(6) }}
+```
+
+**Output**
+
+```jinja
+one
+      two
+      three
+```
+
+Change default indentation to 6 spaces and indent the first line:
+
+**Input**
+
+```jinja
+{{ "one\ntwo\nthree" | indent(6, true) }}
+```
+
+**Output**
+
+```jinja
+      one
+      two
+      three
 ```
 
 ### int
@@ -1112,7 +1144,7 @@ respectively. The base is ignored for decimal numbers and non-string values.
 
 ### join
 
-You can join items together like so:
+Return a string which is the concatenation of the strings in a sequence.
 
 **Input**
 
@@ -1126,7 +1158,9 @@ You can join items together like so:
 ```jinja
 123
 ```
-Or you can join like so:
+
+The separator between elements is an empty string by default.
+It can be defined with an optional parameter:
 
 **Input**
 
@@ -1141,7 +1175,7 @@ Or you can join like so:
 foo,bar,bear
 ```
 
-Or you can join on attributes of an object like so:
+This will also work for arrays:
 
 **Input**
 
@@ -1241,6 +1275,7 @@ This will change everytime the page is refreshed.
 ```
 
 **Output**
+
 A random value between 1-9 (inclusive).
 
 
@@ -1253,7 +1288,7 @@ If no test is specified, the attribute’s value will be evaluated as a boolean.
 
 This would test for false Outputs and then return the length of that Output.
 
-This is the opposite of ```selectattr```
+This is the opposite of ```selectattr``` since it is the only falsy statement:
 
 **Input**
 
@@ -1267,8 +1302,6 @@ This is the opposite of ```selectattr```
 ```jinja
 1
 ```
-
-Since it is the only falsy statement.
 
 ### replace (the first argument can take a JS regular expression)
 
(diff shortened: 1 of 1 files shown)


docs/templating.md
```diff
@@ -892,41 +892,749 @@ otherwise.
...
+```jinja
+one
+    two
+    three
+```
+
+**Input**
+
+```jinja
+{{ "one\ntwo\nthree" | indent }}
```
carljm: missing parameter here?

docs/templating.md
```diff
@@ -892,41 +892,749 @@ otherwise.
...
+
+### wordcount
+
+Output the number of words in the item
+
+**Input**
+
+```
+{{ item.description | wordcount }}
+```
```
carljm: missing outcome here?

docs/templating.md
```diff
@@ -892,41 +892,749 @@ otherwise.
...
+{{ "foo http://www.example.com/ bar" | urlize | safe }}
+```
+
+**Outcome**
+
+```jinja
+foo <a href="http://www.example.com/">http://www.example.com/</a> bar
+```
+
+You can also truncate URL text
```
carljm: missing colon at end of line?

docs/templating.md
```diff
@@ -892,41 +892,749 @@ otherwise.
...
+
+**Outcome**
+
+```jinja
+%26
+```
+
+### urlize
+
+Converts URLs in plain text into clickable links.
```
carljm: `Convert` instead of `Converts`, for consistent mood throughout

docs/templating.md
```diff
@@ -892,41 +892,749 @@ otherwise.
...
+
+**Outcome**
+
+```jinja
+FOO
+```
+
+### urlencode
+
+URL Encode the string
```
carljm: Missing full stop?

docs/templating.md
```diff
@@ -892,41 +892,749 @@ otherwise.
...
+
+```jinja
+{{ "  foo " | trim }}
+```
+
+### truncate
+
+Return a truncated copy of the string. The length is specified with the first parameter which defaults to 255. If the second parameter is true the filter will cut the text at length. Otherwise it will discard the last word. If the text was in fact truncated it will append an ellipsis sign ("..."). If you want a different ellipsis sign than "..." you can specify it using the third parameter.
+
+Truncate to 3 characters
```
carljm: Colon at end of line?

docs/templating.md
```diff
@@ -892,41 +892,749 @@ otherwise.
...
+
+**Input**
+
+```jinja
+{{ "foo bar" | truncate(3) }}
+```
+
+Truncate to 6 characters and append "?" to the end
```
carljm: Colon at end of line?

docs/templating.md
```diff
@@ -892,41 +892,749 @@ otherwise.
...
+
+### trim
+
+Strip leading and trailing whitespace.
+
+**Input**
+
+```jinja
+{{ "  foo " | trim }}
+```
```
carljm: Missing output?

docs/templating.md
```diff
@@ -892,41 +892,749 @@ otherwise.
...
+
+**Outcome**
+
+```jinja
+1,2,3,4,
+```
+
+### sum
+
+The total of the items in the array
```
carljm: "Output the sum of items in the array."

docs/templating.md
```diff
@@ -892,41 +892,749 @@ otherwise.
...
+**Outcome**
+
+```jinja
+2
+```
+
+
+### slice
+
+Slice an iterator and return a list of lists containing those items. Useful if you want to create a div containing three ul tags that represent columns:
```
carljm: Hard wrap at 79 columns

docs/templating.md
```diff
@@ -892,41 +892,749 @@ otherwise.
...
+```
+
+### safe
+
+Mark the value as safe which means that in an environment with automatic escaping enabled this variable will not be escaped.
+
+
+### selectattr (only the single-argument form)
+
+Filters a sequence of objects by applying a test to the specified attribute of each object, and only selecting the objects with the test succeeding.
```
carljm: Hard wrap

docs/templating.md
```diff
@@ -892,41 +892,749 @@ otherwise.
...
+
+**Outcome**
+
+```jinja
+4.1235
+```
+
+### safe
+
+Mark the value as safe which means that in an environment with automatic escaping enabled this variable will not be escaped.
```
carljm: hard wrap

docs/templating.md
```diff
@@ -892,41 +892,749 @@ otherwise.
...
+{{ 4 | round(0, "floor")
+```
+
+**Outcome**
+
+```jinja
+4
+```
+
+Or we can choose how many digits to round up.
```
carljm: colon?

docs/templating.md
```diff
@@ -892,41 +892,749 @@ otherwise.
...
+```
+Note in this instance the required quote marks surrounding the list.
+
+**Outcome**
+
+```jinja
+xxabbbccc
+```
+
+Finally, you can also look for patterns in your list to replace
```
carljm: colon

docs/templating.md
```diff
@@ -892,41 +892,749 @@ otherwise.
...
+```
+
+**Outcome**
+
+```jinja
+.l.e.t.t.e.r.s.
+
+```
+
+We can also replace every instance of an item up to a given number (item to be replaced, item replacement, number to be replaced)
```
carljm: hard wrap and colon

carljm:
Looks great! Made some minor comments - please hard wrap to 79 columns throughout, check consistency of punctuation and mood (use imperative mood: "Convert the thing" not "converts the thing"), check places where output example is missing. Also maybe use "Output" instead of "Outcome"? Thanks!

vipickering:
Thanks for the feedback, I'll make the updates and re-submit

carljm:
(no need to create a new PR, in case that's what you meant by "re-submit"; you can just push new commits to your same branch and the PR will automatically pick them up)

vipickering:
Thanks will do!

vipickering:
@carljm I think I've got all the issues. When you get a chance please do another pass and let me know if there is anything outstanding.

thanks :-)

docs/templating.md
```diff
@@ -892,41 +892,816 @@ otherwise.
...
+
+**Output**
+
+```jinja
+3
+```
+
+### batch
+
+Batch returns a list of lists with the given number of items:
```
carljm: "Return a list of lists...": imperative mood, no need to repeat name of filter

docs/templating.md
```diff
@@ -892,41 +892,816 @@ otherwise.
...
+
+**Output**
+
+```jinja
+12-34-56
+```
+
+### capitalize
+
+Makes the first letter uppercase, the rest lower case:
```
carljm: Make

docs/templating.md
```diff
@@ -892,41 +892,816 @@ otherwise.
...
+**Output**
+
+```jinja
+This is a test
+```
+
+
+### center
+
+Centers the value in a field of a given width:
```
carljm: Center

docs/templating.md
```diff
@@ -892,41 +892,816 @@ otherwise.
...
+```
+
+**Output**
+
+A random value between 1-9 (inclusive).
+
+
+### rejectattr (only the single-argument form)
+
+This is the opposite of ```selectattr``` filter.
```
carljm: Put this paragraph second.

docs/templating.md
```diff
@@ -892,41 +892,816 @@ otherwise.
...
+
+Filters a sequence of objects by applying a test to the specified attribute
```
carljm: Filter

docs/templating.md
```diff
@@ -892,41 +892,816 @@ otherwise.
...
+{{ foods | rejectattr("tasty") | length }}
+```
+
+**Output**
+
+```jinja
+1
+```
+
+### replace (the first argument can take a JS regular expression)
```
carljm: why does this parenthetical belong in the header?

docs/templating.md
```diff
@@ -892,41 +892,816 @@ otherwise.
...
+
+**Output**
+
+```jinja
+foo <a href="http://www.example.com/">http://www.example.com/</a> bar
+```
+
+### selectattr (only the single-argument form)
+
+This is the opposite to ```rejectattr```.
```
carljm: Put this second.

docs/templating.md
```diff
@@ -892,41 +892,816 @@ otherwise.
...
+
+Filters a sequence of objects by applying a test to the specified attribute
```
carljm: Filter

docs/templating.md
```diff
@@ -892,41 +892,816 @@ otherwise.
...
+    <ul class="column-3">
+      <li>7</li>
+      <li>8</li>
+      <li>9</li>
+    </ul>
+</div>
+```
+
+### string
+Make a string unicode if it isn’t already:
```
carljm: Add blank line here?

carljm: Was this description copied from the Jinja docs? I don't think it really applies to nunjucks; JS doesn't have a `unicode` data type. Maybe "convert an object to a string"?

carljm:
Looking good! Seems pretty close, I marked a few remaining issues

docs/templating.md
vipickering: Yes sorry. I initially copied all the Jinja descriptions when I ported it across to makes sure I had everything and a general gist of what it did. I've updated it to be more in line with JS as suggested.

docs/templating.md
vipickering: Changed to 

> Return a list of lists with the given number of items:

docs/templating.md
vipickering: updated

docs/templating.md
vipickering: updated

docs/templating.md
vipickering: moved to second paragraph

docs/templating.md
vipickering: updated

docs/templating.md
vipickering: It doesn't. Good spot, I've removed it.

docs/templating.md
vipickering: moved to second paragraph.

docs/templating.md
vipickering: updated

vipickering:
Thanks for all the great feedback. I've updated the docs as suggested.

Let me know if you spot anymore

carljm:
Thank you! Merging.

PR #813, opened 2016-08-21
Add support for finding object length (#283)

### LINKED ISSUE #283: `length` filter should also handle objects
Make `length` return the number of a keys an object has.

kevinschaul:
In length builtin filter. For objects, "length" is the number of keys
the object has.

Fixes #283

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,8 @@ Changelog
 2.x (unreleased)
 ----------------
 
+* Add support for finding an object's "length" in length filter.
+
 * Ensure that precompiling on Windows still outputs POSIX-style path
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -30,6 +30,8 @@ Changelog
 2.x (unreleased)
 ----------------
 
+* Add support for finding an object's "length" in length filter.
```
carljm: Can you add a link to the PR here, like in the other entries?

carljm:
Looks great, thank you! Requested one minor addition to changelog, otherwise this is merge-ready.

kevinschaul:
Awesome! I just added that link.

PR #825, opened 2016-08-31
Broadened Windows precompile support to include strings

kim3er:
I initially created #761 to deal with incorrectly formatted paths, when precompiling in Windows. The original pull request only dealt with directories, this change also deals with strings.

### DIFF
--- a/tests/precompile.js
+++ b/tests/precompile.js
@@ -31,6 +31,20 @@
 
                 expect(fileName).to.not.contain('\\');
             });
+
+            it('should return *NIX path seperators, when name is passed as option', function() {
+                var fileName;
+
+                precompile('<span>test</span>', {
+                    name: 'path\\to\\file.j2',
+                    isString: true,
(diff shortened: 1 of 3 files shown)


tests/precompile.js
```diff
@@ -31,6 +31,20 @@
...
+
+                precompile('<span>test</span>', {
+                    name: 'path\\to\\file.j2',
+                    isString: true,
+                    wrapper: function(templates) {
+                        fileName = templates[0].name;
+                    }
+                });
+
+                expect(fileName).to.not.contain('\\');
```
carljm: why not make a stronger assertion about what the filename should be?

kim3er:
Updated with clearer assertions on resulting names.

carljm:
Can you add a changelog entry (in the 2.x section) with a link to this PR, like the existing entries? Thanks!

kim3er:
Done. :)

PR #826, opened 2016-08-31
Add elseif case ("alias" of elif) to parseIf

kswedberg:
This commit adds an `elseif` case that falls through to `elif`. Having `elseif` provides consistency between nunjucks and twig so the same templates can be used with either one.


carljm:
Can you add a test for this?

kswedberg:
Sure, I'd be happy to do that. Any suggestions for what exactly to test? I was thinking I'd replicate an `elif` test, bit the only test I see for `elif` in /tests/parser.js is this one:

```js
expect(function() {
    parser.parse('hello {% if sdf %} data');
}).to.throwException(/expected elif, else, or endif/);
```

carljm:
Heh, looks like the existing testing of `elif` leaves something to be desired. I'd just add a compiler test (these are in practice our end-to-end functional tests) showing that if/elseif/endif works as expected. And if you're up for adding one for `elif` while you're at it, that'd be great :-)

carljm:
(by "works as expected" I just mean something simple like `{% if foo == 1 %}one{% elseif foo == 2 %}two{% endif %}` and pass in a few different values for `foo` in the context, assert the template output.

carljm:
Tests look great, thanks! And sorry I forgot to mention this in the first round, but would you also be willing to add a changelog entry (in the 2.x section), following the pattern of the existing entries? Thanks!

kswedberg:
Sure thing. 

Also, not sure if you want docs updated or not, but I could change:

> You can specify alternate conditions with `elif` and `else`

to: 

> You can specify alternate conditions with `elif` (or `elseif`, which is simply an alias of `elif`) and `else`

carljm:
Sure, please do update the docs as well.

kswedberg:
Ok, I think I have everything in there, and I hope I didn't make too much of a mess out of the merge I just did. I thanked myself in the changelog, which was weird, but it fit the pattern, so what the heck.

carljm:
Looks great, thank you!

naterkane:
@carljm @kswedberg (Hi Karl, long time to talk!)
do either of you know if this has been released yet? I just stumbled on this PR as I was trying to see why elseif wasn't working. 

I'm currently using   
* "nunjucks": "2.5.x"
* "nunjucks-markdown": "2.0.x"

I'm not sure that I want to bump any dependencies at the moment.

kswedberg:
Hey @naterkane ! Hope you're well.
According to the commit history, `elseif` was included in the `2.5.0` release. (https://github.com/mozilla/nunjucks/commit/0ed30890437498493eb346220791ac314e616d1e#diff-48449cc0d8ca4858eddbb24677cf304eR3078). I've been using it successfully since then. 

Are you sure you're using 2.5+? If you go into `node_modules/nunjucks/package.json`, what's the version property?

naterkane:
yes, I'm currently using `2.5.2`

```
"version": "2.5.2"
```

kswedberg:
Huh. that's strange. So, `elif` works, but `elseif` doesn't? if there's any way I can take a look at your setup, I'd be happy to help investigate

naterkane:
Sure. Template files in github.com/naterkane/raml2html there's a single use
of `elif`

- sent from my potato

kswedberg:
ok, cool. I cloned the repo and `npm install`ed. Changed `elif` to `elseif` in `lib/resource.nunjucks` and then ran `node script` from inside `examples` directory. No errors. 

Is there some other way I should be testing this? Tried running `npm run examples`, but that didn't work for other reasons (both `elif` and `elseif`)

naterkane:
It uses a git submodule (at the moment) going to refactor to use a npm
module to provide the RAML examples tomorrow.

- sent from my potato

kswedberg:
ah, ok. I got the submodule in there and ran `npm run examples`, still with no errors. everything looks good. if there's somewhere in particular where you're seeing things break, more information would be helpful. so far, though, I can't seem to reproduce the problem.

naterkane:
Very strange. I'll look more into it tonight. If you want to PR your
`elseif` change that would be great.

- sent from my potato

On Jan 7, 2017 12:56 PM, "Karl Swedberg" <notifications@github.com> wrote:

ah, ok. I got the submodule in there and ran npm run examples, still with
no errors. everything looks good. if there's somewhere in particular where
you're seeing things break, more information would be helpful. so far,
though, I can't seem to reproduce the problem.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/mozilla/nunjucks/pull/826#issuecomment-271099067>, or mute
the thread

Issue #777, opened 2016-06-07
Programmatic templates precompiling


kolesnikova-tavi:
I'm trying to use gulp to precompile templates. My plan is:
1) Precompile templates to some files;
2) Require;
3) Use render.

I'm doing something like this (gulpfile.js):
```
    return merge(dirs.map(function(dir) {
        return gulp.src(path.join(dir, '/*.html'))
            .pipe(pipeError('Precompile Folder Templates'))
            .pipe(gulpNunjucks.precompile({
                path: 'src/templates/modules',
                options: {
                    noCache: true
                }
                context: {
                    SOME_VARS: '',
                    SOME_FUNCTIONS: function() {
                    }
                },
                extendEnv: function(env) {
                    // ...
                }
            }))
            .pipe(concat('templates.js'))
            .pipe(gulp.dest(path.dirname(dir)));
        }));
```

gulpNunjucks.precompile does this:
```
// ...
settings.loader = new nunjucks.FileSystemLoader(settings.path);
settings.env = new nunjucks.Environment(settings.loader, settings.options);
settings.extendEnv(settings.env);
// ...
let output = 'exports["<%= file.stem %>"]=<%= template %>;';
// ...
// TODO: almost what I need (*)
let result = nunjucks.precompileString(file.contents.toString(), {
    env: settings.env,
    name: file.stem,
    context: settings.context  // TODO: need pass context (**)
});
file.contents = new Buffer(result);
file.path = gutil.replaceExtension(file.path, '.js');
```

Then in my node project js-file:
`let templates = require('./moduleName/templates');`

And I need something like this:
```
// TODO: need a way to render (***)
templates.loginForm.render(context);
templates.regForm.render(context);
```	
(\*) This way to use precompiled files with tag script. How can I get function without using window object?
(\*\*) How can I pass context?
(\*\*\*) I can't see how use precompiled templates.
	
Is any way?
Thanks!

rhengles:
Please take a look at this gist I created:

https://gist.github.com/rhengles/2a4b4df5aedb063f6a8d79029f66f47b

Maybe we could use this issue to improve the documentation in this regard.

carljm:
PR for improved documentation would be most welcome!

kolesnikova-tavi:
It works! Thanks a lot!
Can I pass variables to precompile method?

rhengles:
@kolesnikova-tavi No. You can add filters and extensions [1], but the context variables are only used on the render method.

[...]

PR #832, opened 2016-09-04
Add to precompile documentation

olsonpm:
Just had to learn how this works so I figured i'd create a PR.

My appended documentation is convoluted, but so is the code exposing that functionality.  I'd almost rather remove the explanation of the default return value and just advise against its usage.  The returned string is too specific to provide generic use.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -629,7 +629,7 @@ everything in sync.
 precompile
 nunjucks.precompile(path, [opts])
 
-Precompile a file or directory at **path**. **opts** is a hash with any of the following options:
+Synchronously precompile a file or directory at **path**. **opts** is a hash with any of the following options:
 
 * **name**: name of the template, when compiling a string (required)
     or a file (optional, defaults to **path**). names are
@@ -644,6 +644,12 @@ Precompile a file or directory at **path**. **opts** is a hash with any of the f
         * **name**: name of the template
         * **template**: string source of the precompiled template in javascript
     * **opts**: object of all the above options
+
+This function returns the output of `opts.wrapper` if passed, defaulting to a string encompassing executable javascript
+that will assign each precompiled template to `window.nunjucksPrecompiled[<templateName>]`.  **asFunction**
+modifies this slightly by also returning a function that renders each template.  
+
+Note: The default return value is not meant for general api usage.  You should instead pass in `opts.wrapper`
 
 ```js
 var env = new nunjucks.Environment();


rhengles:
That's nice! I'd just like to use this opportunity to ask that, if possible, could you use this PR to solve this related issue (#777) ? I think that this section of the documentation lacks some examples.

If you don't feel confortable with that, there's no problem. I just wanted to point out that issue.

Thanks!

olsonpm:
Well today I need to tackle precompilation via the nunjucks api for templates to be used on the browser.  I don't mind adding an example, but I won't have time to add multiple for different use-cases.  I'm also new to nunjucks, and after reading through some issues regarding browser packaging - I couldn't gather how people most commonly consume nunjucks client-side.  Seems that most just add a `<script src="path/to/nunjucks.js">`.  I'm using webpack.  What sort of consumption should the example have in mind?

olsonpm:
fyi - this is the implementation I landed on.  It's hacky, but relying on `window.nunjucksPrecompiled` is more hacky.

```js
const nunjucksSlim = require('nunjucks/browser/nunjucks-slim')
  // 'precompiled-templates' is the cjs compatable output from a modified version of your wrapper module
  , precompiledTemplates = require('./precompiled-templates')
...
const nunjucks = initiateNunjucks();
...
function initiateNunjucks() {
  const env = nunjucksSlim.configure();
  env.loaders.length = 0;
  const precompiledTemplatesLoader = new nunjucksSlim.PrecompiledLoader(precompiledTemplates);
  precompiledTemplatesLoader.cache = {};
  env.loaders.push(precompiledTemplatesLoader);
  
  return env;
}
```

rhengles:
@olsonpm Forgive my curiosity, but what is your modification to the cjs wrapper module?

Yes, I didn't like the global variable either, that's why I created the wrapper option.

olsonpm:
No problem - just added an option for a "name-mapper" function.  I didn't want ".njk" anywhere in my browser code.  Personal preference

docs/api.md
```diff
@@ -645,6 +645,12 @@ Precompile a file or directory at **path**. **opts** is a hash with any of the f
         * **template**: string source of the precompiled template in javascript
     * **opts**: object of all the above options
 
+This function returns the output of `opts.wrapper` if passed, defaulting to a string encompassing executable javascript
+that will assign each precompiled template to `window.nunjucksPrecompiled[<templateName>]`.  **asFunction**
+modifies this slightly by also returning a function that renders each template.  
+
+Note: The default return value is not meant for general api usage.  You should instead pass in `opts.wrapper`
```
ogonkov: Return value of precompiled template?

olsonpm:
I haven't used nunjucks in a long time.  I won't be able to finish this PR.

PR #836, opened 2016-09-07
fix autoescape for non-string values; fixes #835

### LINKED ISSUE #835: Bypass autoescape / XSS
The following string works as expected: 

``` javascript
var res = nunjucks.renderString('Hello {{ username }}', { username: '<s>Matt</s>' });
console.log(res); //Hello &lt;s&gt;Matt&lt;/s&gt;
```

If however the variable passed to the template is an array autoescape does nothing: 

``` javascript
var res = nunjucks.renderString('Hello {{ username }}', { username: ['<s>Matt</s>'] });
console.log(res); // Hello <s>Matt</s>
```

This looks to be intentional https://github.com/mozilla/nunjucks/blob/master/src/runtime.js#L209 
However if a autoescape is on any variable that is rendered in a {{ }} block and appended to the output I would expect it to escaped no matter the type.  If the var is just going to be part of a string concat anyway why not toString first then escape?  

In express / Koa / (anything else using qs or body-parser) is is trivial to coerce query params types. See the following simple example in express: 

``` javascript
var app = express();
var env = nunjucks.configure('views', {
    autoescape: true,
    express: app
});

app.get('/', function(req, res) {     
    // This should be fine autoescape is on...
    res.send(nunjucks.renderString('Hello {{ username }}', { username: req.query.name }));
    //res.render('index.html', { username: req.query.name }); 
});
```
`http://127.0.0.1:3000/?name[]=<script>alert(1)</script>matt`

I created a more detailed writeup and example app at:  https://github.com/matt-/nunjucks_test

devoidfury:
#835

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ Changelog
   macro. Merge of [#791](https://github.com/mozilla/nunjucks/pull/791).
 
 * Fix potential cast-related XSS vulnerability in autoescape mode, and with `escape` filter.
+  Thanks Matt Austin for the report and Thomas Hunkapiller for the fix.
   [#836](https://github.com/mozilla/nunjucks/pull/836)
 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,9 @@ Changelog
 * Fix handling of macro arg with default value which shares a name with another
   macro. Merge of [#791](https://github.com/mozilla/nunjucks/pull/791).
 
+* Fix potential cast-related XSS vulnerability in autoescape mode.
+  [#836](https://github.com/mozilla/nunjucks/pull/836)
+
 
 2.x (unreleased)
 ----------------
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,7 @@ Changelog
 * Fix handling of macro arg with default value which shares a name with another
   macro. Merge of [#791](https://github.com/mozilla/nunjucks/pull/791).
 
-* Fix potential cast-related XSS vulnerability in autoescape mode.
+* Fix potential cast-related XSS vulnerability in autoescape mode, and with `escape` filter.
   [#836](https://github.com/mozilla/nunjucks/pull/836)
 
 
--- a/src/filters.js
+++ b/src/filters.js
@@ -116,10 +116,10 @@ var filters = {
     },
 
     escape: function(str) {
-        if(typeof str === 'string') {
-            return r.markSafe(lib.escape(str));
+        if(str instanceof r.SafeString) {
+            return str;
         }
-        return str;
+        return r.markSafe(lib.escape(str.toString()));
     },
 
     safe: function(str) {
--- a/src/runtime.js
+++ b/src/runtime.js
@@ -206,8 +206,8 @@ function markSafe(val) {
 function suppressValue(val, autoescape) {
     val = (val !== undefined && val !== null) ? val : '';
 
-    if(autoescape && typeof val === 'string') {
-        val = lib.escape(val);
+    if(autoescape && !(val instanceof SafeString)) {
+        val = lib.escape(''+val);
     }
 
     return val;
--- a/src/runtime.js
+++ b/src/runtime.js
@@ -207,7 +207,7 @@ function suppressValue(val, autoescape) {
     val = (val !== undefined && val !== null) ? val : '';
 
     if(autoescape && !(val instanceof SafeString)) {
-        val = lib.escape(''+val);
+        val = lib.escape(val.toString());
     }
 
     return val;
--- a/tests/compiler.js
+++ b/tests/compiler.js
@@ -1395,6 +1395,14 @@
                 expect(res).to.be('&><\'"');
             });
 
+            render('{{ foo }}', { foo: ['<p>foo</p>']}, { autoescape: true }, function(err, res) {
+                expect(res).to.be('&lt;p&gt;foo&lt;/p&gt;');
+            });
+
+            render('{{ foo }}', { foo: {toString: function() {return '<p>foo</p>'}}}, { autoescape: true }, function(err, res) {
+                expect(res).to.be('&lt;p&gt;foo&lt;/p&gt;');
+            });
+
             finish(done);
         });
 
--- a/tests/filters.js
+++ b/tests/filters.js
@@ -110,6 +110,15 @@
             finish(done);
         });
 
+        it('should work with non-string values', function(done) {
+            var res1 = render('{{ foo | escape }}', {foo: ['<html>']}, { autoescape: false });
+            expect(res1).to.be('&lt;html&gt;');
+
+            var res2 = render('{{ foo | escape }}', {foo: {toString: function() { return '<html>'; }}}, { autoescape: false });
+            expect(res2).to.be('&lt;html&gt;');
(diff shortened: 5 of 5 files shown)


devoidfury:
Imo, a bugfix for the issue is release asap material (even if it's not this one). Maybe submit a CVE?

@SamyPesse @jlongster @vecmezoni  @carljm @jbmoelker

src/runtime.js
```diff
@@ -206,8 +206,8 @@ function markSafe(val) {
 function suppressValue(val, autoescape) {
     val = (val !== undefined && val !== null) ? val : '';
 
-    if(autoescape && typeof val === 'string') {
-        val = lib.escape(val);
+    if(autoescape && !(val instanceof SafeString)) {
+        val = lib.escape(''+val);
```
vecmezoni: `val.toString()` will be clearer

jbmoelker:
@devoidfury thanks for this! Looks solid. Could you update the `CHANGELOG.md` with the fix? For me that's all that's left before merging this PR.

I don't know if we need to submit a CVE. I understand the risk of this bug. But I don't know the protocol or procedure.

@carljm what's the procedure to release this? I've read `MAINTENANCE.md`, but I don't quite understand how the `2.x` and `3.x` branches are managed. Seeing the [unreleased 2.x items](https://github.com/mozilla/nunjucks/blob/master/CHANGELOG.md#2x-unreleased) I imagine we should release `2.5.0`. Also I think @vecmezoni and I as 1 day old maintainers are able to create releases on GitHub, but are unable to publish a new version to npm.

vecmezoni:
Probably would be better to publish 2.4.3 and 2.5.0 to fix vulnerability, so users with all the cases of package.json versioning `~ ^` will receive fix. Just need to check is there anything to bump a minor version or we might bump just patch (2.4.3)

jbmoelker:
@vecmezoni well there are multiple [unreleased features on 2.x](https://github.com/mozilla/nunjucks/blob/master/CHANGELOG.md#2x-unreleased), so that's why the minor bump. I agree that normally we'd merge this PR and release patch. But I'm not sure where to merge onto for that patch release.

vecmezoni:
We can make separate branch from 2.4.2 version, apply this patch and release 2.4.3. Then we can apply this patch to 2.x and decide to release or not to release 2.5. I've asked @carljm to give me access to npm, waiting.

vecmezoni:
In this case people with `"nunjucks": "^2.4"` will get this update. Also we can add deprecation warning for everything less than 2.4.2 https://docs.npmjs.com/cli/deprecate

vecmezoni:
@devoidfury add note to `CHANGELOG.md` and i'm going to merge this and prepare for release and deprecation note.

matt-:
Should the escape filter also do the same?

devoidfury:
Yes, thanks @matt- -- added another commit to fix this.

CHANGELOG.md
```diff
@@ -26,6 +26,9 @@ Changelog
 * Fix handling of macro arg with default value which shares a name with another
   macro. Merge of [#791](https://github.com/mozilla/nunjucks/pull/791).
 
+* Fix potential cast-related XSS vulnerability in autoescape mode, and with `escape` filter.
```
jbmoelker: Can be merged without, but credits where due: "Thanks Matt Austin and Thomas Hunkapiller.".

vecmezoni:
👍  Merging this to master, then i'll create branch 2.4.3 from 2.4.2 release, backport this patch and rebase 2.x branch on 2.4.3.

carljm:
@vecmezoni sounds good, but I'd avoid rebasing the `2.x` branch. It's public, that could really mess up anyone else tracking it. just merge from your new 2.4.3 branch into 2.x instead.

vecmezoni:
* Released 2.4.3
* Added deprecation note (i'm not sure that vulnerability is critical, but anyways)
```
> npm i nunjucks@2.4.2

npm WARN deprecated nunjucks@2.4.2: potential XSS vulnerability in autoescape mode, 
and with escape filter was fixed in v2.4.3
```
* Backported patch to 2.x, tomorrow i will review and test branch and issue 2.5.0 release.

dlmr:
Seems like this change broke the use of the `safe` filter. 

```
{{ value | safe }}
```

With `2.4.3` this will be escaped where I thought it should not be that. Could be that I have used it incorrectly in my application and that this was never supposed to work but wanted to give a heads up if this was not the case.

devoidfury:
@dlmr Could you provide a simple test case here?

dlmr:
@devoidfury I think I found the problem when trying to create a minimal example.

I was using [React Helmet](https://github.com/nfl/react-helmet) but did not use the `toString()` function on what I got back from it which means that the safe filter did nothing in that case and then when rendering the escape did not work, since it was not a string, and I got the unescaped output as I wanted.

So I was actually this bug in my code. The more correct thing is of course to use `toString` and in the process be more correct. 

Thanks for the quick reply and the fix in the first place! Sorry to blame the PR for this! 😄

nelsonpecora:
Haha thanks @dimr for catching this. I just hit this with [an old filter we were using](https://github.com/nymag/nunjucks-filters/blob/master/filters/files/includeFile.js) for embedding optional svgs. It definitely woke me up this morning!

screen shot 2016-09-08 at 11 58 41 am

If I'm parsing the tests in this PR correctly, I should be doing something like this in my filter, correct?

```js
'use strict';
var fs = require('fs');

module.exports = function (path) {
  try {
    return {
      toString: function () {
        return fs.readFileSync(path);
      }
    };
  } catch () {
    return '';
  }
};
```

devoidfury:
@vecmezoni @carljm should we modify the SafeString constructor to automatically do a string conversion? Right now, it's no-op (identity) on non-strings, but after that point it's going to be cast to string anyway, and with autoescape, if it's not a SafeString, it will escape.

fengmk2:
Should we fix this in 1.x?

src/filters.js
```diff
@@ -116,10 +116,10 @@ var filters = {
     },
 
     escape: function(str) {
-        if(typeof str === 'string') {
-            return r.markSafe(lib.escape(str));
+        if(str instanceof r.SafeString) {
+            return str;
         }
-        return str;
+        return r.markSafe(lib.escape(str.toString()));
```
fengmk2: this will throw `TypeError: Cannot read property 'toString' of null` when str is `null` or `undefined`

tests/compiler.js
```diff
@@ -1395,6 +1395,14 @@
                 expect(res).to.be('&><\'"');
             });
 
+            render('{{ foo }}', { foo: ['<p>foo</p>']}, { autoescape: true }, function(err, res) {
```
fengmk2: Add one more test case for `null` value check:

```js
render('{{ foo }}', { foo: null }, { autoescape: true }, function(err, res) {
  // ...
});
```

tests/filters.js
```diff
@@ -110,6 +110,15 @@
             finish(done);
         });
 
+        it('should work with non-string values', function(done) {
+            var res1 = render('{{ foo | escape }}', {foo: ['<html>']}, { autoescape: false });
```
fengmk2: Add one more test case for null value check.

src/filters.js
atian25: https://github.com/mozilla/nunjucks/pull/843

PR #843, opened 2016-09-12
fix: undefined escape at #836

atian25:
relate #843
https://github.com/mozilla/nunjucks/pull/836/files#r78286483

### DIFF
--- a/src/filters.js
+++ b/src/filters.js
@@ -119,10 +119,12 @@ var filters = {
         if(str instanceof r.SafeString) {
             return str;
         }
+        str = (str === null || str === undefined) ? '' : str;
         return r.markSafe(lib.escape(str.toString()));
     },
 
     safe: function(str) {
+        str = (str === null || str === undefined) ? '' : str;
         return r.markSafe(str);
     },
 
--- a/tests/compiler.js
+++ b/tests/compiler.js
@@ -1309,8 +1309,20 @@
                 expect(res).to.be('&><\'"');
             });
 
+            render('{{ foo }}', { foo: null }, { autoescape: true }, function (err, res) {
+                expect(res).to.be('');
+            });
+
+            render('{{ foo | safe }}', { foo: null }, { autoescape: true }, function (err, res) {
+                expect(res).to.be('');
+            });
+
             render('{{ foo }}', { foo: ['<p>foo</p>']}, { autoescape: true }, function(err, res) {
                 expect(res).to.be('&lt;p&gt;foo&lt;/p&gt;');
+            });
+
+            render('{{ foo | safe }}', { foo: '<p>foo</p>' }, { autoescape: true }, function (err, res) {
+                expect(res).to.be('<p>foo</p>');
             });
 
             render('{{ foo }}', { foo: {toString: function() {return '<p>foo</p>'}}}, { autoescape: true }, function(err, res) {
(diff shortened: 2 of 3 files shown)


src/filters.js
```diff
@@ -119,6 +119,7 @@ var filters = {
         if(str instanceof r.SafeString) {
             return str;
         }
+        if (str == null) str = '';
```
vecmezoni: Instead of this, please use https://github.com/atian25/nunjucks/blob/d75a9993c79fb513e9e651bc938f5555213cb2a7/src/filters.js#L6

atian25: `if(value === null || value === undefined || value === false) {`

`render('{{ foo }}', { foo: false })` should render `false` or `empty string` ?

fengmk2:
@atian25 should fix `safe` filter bug too.

tests/compiler.js
```diff
@@ -1395,6 +1395,10 @@
                 expect(res).to.be('&><\'"');
             });
 
+            render('{{ foo }}', { foo: null }, { autoescape: true }, function (err, res) {
+                expect(res).to.be('');
+            });
```
fengmk2: add `render('{{ foo | safe }}', { foo: '<html>' }, { autoescape: true }` test for safe filter.

atian25: done

atian25:
@vecmezoni updated.

question: 

`if(value === null || value === undefined || value === false) {`

`render('{{ foo }}', { foo: false })` should render `false` or `empty string` ?

fengmk2:
boolean value should follow `toString()` logic.

atian25:
@fengmk2  agree. updated.

src/filters.js
```diff
@@ -119,10 +119,12 @@ var filters = {
         if(str instanceof r.SafeString) {
             return str;
         }
+        str = (str === null || str === undefined) ? '' : str;
```
vecmezoni: Could you please fix normalize function and remove false from there and use normalize here.
I don't see any idea behind it

atian25: maybe another PR.  This one is focus at quick bug fix.
I'm afraid it will cause BREAK CHANGE.

vecmezoni: ok, i'll merge it to 2.x, please prepare pull request for master

vecmezoni: could you rebase this branch on 2.x?

atian25: ok, wait a moment

atian25: @vecmezoni plz review again. The ci is fail, but local test is ok, could you trigger it again?

vecmezoni: I can't trigger CI, you can change commit hash and push it with force. It would trigger Appveyour

atian25: @vecmezoni done, travis is pass.

vecmezoni:
Merged. Waiting for `master` branch fix.

vecmezoni:
2.5.1 released, cherry-picked this patch to master.

src/filters.js
```diff
@@ -119,10 +119,12 @@ var filters = {
...
         return r.markSafe(lib.escape(str.toString()));
     },
 
     safe: function(str) {
+        str = (str === null || str === undefined) ? '' : str;
```
atian25: @fengmk2 yes..  missing sth. https://github.com/node-modules/nunjucks/pull/3/files#diff-9f47adb6bfb92783e65552d7436a5cdcR108

Issue #842, opened 2016-09-12
In the latest Node docs code output is being displayed with a red rectangle.


vipickering:
I'll pick this up today, but keeping track here in case anyone has input.

<img width="652" alt="screen shot 2016-09-12 at 08 29 07" src="https://cloud.githubusercontent.com/assets/1380497/18427942/04d0d9d2-78c3-11e6-8dbe-886f5dc2f066.png">

This is happening because we are using "Jinja" as our code highlighting lang. We will need to switch this out for a custom one (or we could do our own!).

The quickest fix for now is to turn off the error output.

vipickering:
Actually investigating this further, it looks like something may have gone wrong in the merge. Some of the tags are also broken.

I'll investigate and fix all the broken bits.

vecmezoni:
Resolved with https://github.com/mozilla/nunjucks/commit/dc004cf10547301159a93aadbe0081be18bc833a

PR #844, opened 2016-09-12
fixed error hightlighting issue. Fixed broken back ticks

vipickering:
Issue #842

I've turned off the code error highlighting for now (via CSS).

I have also fixed a broken back tick reference that was causing the Conditions area of the docs to break and output mangled docs + code.

### DIFF
--- a/docs/_config.yml
+++ b/docs/_config.yml
@@ -1,4 +1,5 @@
 encoding: utf-8
+
 markdown: redcarpet
 redcarpet:
     extensions: [with_toc_data]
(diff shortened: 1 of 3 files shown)


vecmezoni:
I don't think that's good idea to just ignore errors. I'm checking whole process of building docs (jekyll -> pygments.rb -> pygments) to find out why this is happening.

vipickering:
The errors are being highlighted (I think) because the language specified in the backticks is JINJA.

vecmezoni:
Here you can see that same example renders nicely: http://pygments.org/demo/5979977/?style=default

vipickering:
Good spot. I suspected it might be where jinja and Nunjucks had diverged, but clearly that isn't the case.

vecmezoni:
It was enough to add https://github.com/mozilla/nunjucks/commit/dc004cf10547301159a93aadbe0081be18bc833a

Thank's for your help in finding this bug!

PR #849, opened 2016-09-14
fix: safe should call toString

### LINKED ISSUE #847: `{{ arr | safe }}` should not escape
`render('{{ arr | safe }}', { arr: [ '<html>' ] })` => `&lt;html&gt;`

https://github.com/mozilla/nunjucks/blob/master/src/runtime.js#L190

@vecmezoni is this a bug or user should use `{{ arr | join | safe }}` ?

atian25:
close #847

relate: https://github.com/mozilla/nunjucks/pull/843#discussion-diff-78676929

### DIFF
--- a/src/filters.js
+++ b/src/filters.js
@@ -124,8 +124,11 @@ var filters = {
     },
 
     safe: function(str) {
+        if (str instanceof r.SafeString) {
+            return str;
+        }
         str = (str === null || str === undefined) ? '' : str;
-        return r.markSafe(str);
+        return r.markSafe(str.toString());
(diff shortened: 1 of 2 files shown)


vecmezoni:
Well, in jinja2:
```py
import jinja2
jinja2.Template(u'{{ arr | safe }}').render({'arr': ['<html>']}) # == u"['<html>']"
```

But i think we can merge this and then decide what to do with the braces

atian25:
@vecmezoni  thanks, plz merge it and help to publish a patch.

PR #851, opened 2016-09-14
Document purpose

### LINKED ISSUE #837: Set project purpose
I think it helps users / developers, issue reporters and maintainers if the project has a clearly defined purpose. This ensures people know what to expect from the project. They know if it's relevant to report a given issue or if it's "out of scope". And it helps maintainers to judge reported issues and PRs.

Here's my proposal:

* Aim for templating feature parity with Jinja2.
* Aim for templating feature parity with Twig, but only when not conflicting with Jinja2 parity.
* Works in Node >= v0.10 (as defined in [Travis test matrix]).(https://github.com/mozilla/nunjucks/blob/master/.travis.yml#L9).
* Works in all modern browsers (with [ES5 support](http://kangax.github.io/compat-table/es5/)).
* Works in IE8 with [es5-shim](https://github.com/es-shims/es5-shim).
* Keep footprint browser files as small as possible (save on bandwidth, download time).
* Keep performance as fast as possible (see benchmarks).
* Keep maintenance as easy as possible (avoid complexity, automate what we can).

Notes:

* We don't aim for parity of language specific syntax like Python's splice notation `{{ foo[start:end:step] }}`.
* We don't aim for parity of language specific filters like [Twig's PHP date format](http://twig.sensiolabs.org/doc/functions/date.html).

I would like to add this as a section to the [contributing guidelines](../blob/master/CONTRIBUTING.md) and / or [documentation](../blob/master/docs/). But let's first see if we can reach an agreement.

jbmoelker:
Proposed changes:
* Add "project purpose" to contribution guidelines.
* Link to "contribution guidelines" from main README.

Closes #837

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,18 +1,36 @@
-Contributing
-============
+# Contributing
 
-Thanks for your interest in contributing! The advice below will help you get
-your issue fixed / pull request merged.
+Thanks for your interest in contributing! The advice below will help you get your issue fixed / pull request merged.
 
-Please file bugs and send pull requests to the
-[GitHub repository](https://github.com/mozilla/nunjucks/) and
-[issue tracker](https://github.com/mozilla/nunjucks/issues).
+
+## Purpose
+
+Nunjucks has the following purpose:
+
+* Aim for templating feature parity with Jinja2.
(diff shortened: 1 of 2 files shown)


vecmezoni:
👍

PR #871, opened 2016-09-20
fix duplication error in the expressions sections

### LINKED ISSUE #869: Expressions and math section is duplicated in the template docs
- Apparently there was a merge conflict which duplicated the expressions and math section of the template docs.

- I assign myself to this task

JesseE:
- fix for duplication of documentation at the examples section of the expressions documentation.

- apparently the tags weren't escaped properly 

closes #869

### DIFF
--- a/docs/templating.md
+++ b/docs/templating.md
@@ -671,8 +671,8 @@ minus sign (`-`) to the start or end block tag.
 {%- endfor %}
 ```
 
-The exact output of the above would be "12345". The `-%}` strips the whitespace
-right after the tag, and the `{%-` strips the whitespace right before the tag.
+The exact output of the above would be "12345". The `{%-` strips the whitespace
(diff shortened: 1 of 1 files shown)


docs/templating.md
```diff
@@ -671,8 +671,8 @@ minus sign (`-`) to the start or end block tag.
 {%- endfor %}
 ```
 
-The exact output of the above would be "12345". The `-%}` strips the whitespace
-right after the tag, and the `{%-` strips the whitespace right before the tag.
+The exact output of the above would be "12345". The `&#123;%-` strips the whitespace
```
jbmoelker: @JesseE I realise I was the one suggesting escaping special characters to prevent further issues. However it appears the escaped characters are also displayed as such. My apologies. 
Could you, "unescape" them again, and squash your commits? Then I'm okay to merge.

JesseE: @jbmoelker yes will do

PR #883, opened 2016-10-06
Add variables whitespace control

ixth:
## Summary

Proposed change:

Jinja2 supports whitespace control not only in block tags, but also in variables. It's very, very convenient, otherwise you should use awkward workarounds like `{%- if true -%}…{%- endif -%}` &c. I added whitespace control support in variables.
## Checklist

- [x] Proposed change helps towards [_purpose of this project_](https://github.com/mozilla/nunjucks/blob/master/CONTRIBUTING.md#purpose).
- [x] [_Documentation_](https://github.com/mozilla/nunjucks/tree/master/docs/) is added / updated to describe proposed change.
- [x] [_Tests_](https://github.com/mozilla/nunjucks/tree/master/tests) are added / updated to cover proposed change.
- [x] [_Changelog_](https://github.com/mozilla/nunjucks/blob/master/CHANGELOG.md) has an entry for proposed change (if user-facing fix or feature).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,8 @@ Changelog
 * Fix handling of macro arg with default value which shares a name with another
   macro. Merge of [#791](https://github.com/mozilla/nunjucks/pull/791).
 
+* Add variables whitespace control.
+
 
 2.5.2 (Sep 14 2016)
 ----------------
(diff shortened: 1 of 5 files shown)


vecmezoni:
Please supply more test cases, so we would be sure that this won't break anything. For example:
* ```{{- foo -}}```
* ```{{-foo-}}```
* ```{{ -foo }}```
* ```{{- -foo -}}```
* ```{{- foo}}```
* ```{{foo -}}```
* ```{{foo-}}```

vecmezoni:
Rebased and merged

PR #889, opened 2016-10-11
Remove rails/jekyll dependency for docs

jpsc:
## Summary

Proposed change:

I would like to propose that we change the docs from jekyll to using nunjucks itself.
It removes the dependency of rails and it could work as a show of nunjucks power itself.

This is simple POC to see if there is interest, I can work deeper to add languages that I removed for the sake of brevity.


forresst:
@jpsc Personally, I love it:
1) I work a lot on windows and quite complicated to use Jekyll (First point interesting)
2) I think the idea of having a doc that uses nunjuck itself is perfect (Second interesting point)

My opinion may not be enough, but it's beautiful work.

docs/.editorconfig
```diff
@@ -0,0 +1,12 @@
+root = true
+
+[*]
+indent_style = tab
```
vecmezoni: Let's continue to use spaces as indentation:
```
Development/nunjucks [master] » grep -R '\t' ./docs/**/*.md | wc -l
0
Development/nunjucks [master] » grep -R '\t' ./docs/**/*.html | wc -l
0
```

vecmezoni:
This is very cool! I'll take a look at it later today.

docs/_layouts/_subpage.html
```diff
@@ -0,0 +1,39 @@
...
+				<li><a href="templating.html">Templating</a></li>
+				<li><a href="api.html">API</a></li>
+				<li><a href="faq.html">FAQ</a></li>
+			</ul>
+		</div>
+		<div class="toc">
+			<p><strong>{{ page.title }}</strong></p>
+			<ul>
+				{% for heading in toc %}
+					<li>
```
jbmoelker: Suggestion: use `heading.level` as part of CSS class to nest only visually:
```jinja
<li class="toc-level-{{ item.level }}">
  <a href="#{{ item.id }}">{{ item.html | safe }}</a>
</li>
```
used here: https://github.com/jbmoelker/voorhoede-page-formatter/blob/master/src/templates/project.html#L29

docs/api.md
```diff
@@ -781,7 +705,7 @@ This adds a filter `shorten` which returns the first `count`
 characters in a string, with `count` defaulting to 5. Here is how it
 is used:
 
-```jinja
+```html
```
jbmoelker: I'd keep this `jinja` and change it in the formatter (maybe we can also have the formatter accept `nunjucks` and `njk`).

docs/lib/copy-file.js
```diff
@@ -0,0 +1,7 @@
+const promisify = require('bluebird').promisify;
+const copy = promisify(require('ncp').ncp);
+
+module.exports = function (source, destination) {
+	return copy(source, destination)
+		.catch(err => console.log('error copying file', err));
```
jbmoelker: `console.error`?

docs/lib/markdown-to-html.js
```diff
@@ -0,0 +1,11 @@
+'use strict';
+
+const fs = require('fs');
```
jbmoelker: remove `fs`

docs/lib/render-code.js
```diff
@@ -0,0 +1,13 @@
+module.exports = renderCode;
+
+const prism = require('prismjs');
+prism.languages.html = require('./prism-twig'); // use twig (same syntax as nunjucks) as default formatter for html code
```
jbmoelker: you don't need to fork `prism-twig` but you can instead do:

```javascript
const prism = require('prismjs');
require('prismjs/components/prism-json'); // adds `json` to global `prism.languages` instance
require('prismjs/components/prism-twig'); // adds `twig` to global `prism.languages` instance
```
see https://github.com/jbmoelker/jinja-compat-tests/blob/master/lib/nunjucks-filter-prism.js#L1

docs/lib/render-code.js
```diff
@@ -0,0 +1,13 @@
...
+
+function renderCode(code, language) {
+	if (prism.languages.hasOwnProperty(language)) {
```
jbmoelker: if language is `jinja`, `nunjucks` or `njk` set it to `twig` so it's properly highlighted.

docs/scripts/build.js
```diff
@@ -0,0 +1,78 @@
...
+	new nunjucks.FileSystemLoader(baseDir, {
+		noCache: true,
+		watch: false
+	}),
+	{autoescape: false}
+);
+
+const config = {
+	site: {
+		baseurl: ''
```
jbmoelker: determine `baseUrl` per file based on the relative path to root, so if `docs/templating/filters/abs/index.html` and root is `docs/` make sure `baseUrl` becomes `../../../`. If it's always relative it works in any environment.

jpsc: Yes, this is also necessary for the translations, otherwise it won't have the correct paths.

jpsc:
I'm done with the fixed from the review.
I still need someone to fix the CI and integration.

Also would be nice if someone from the translations team had a look.

Going deep into the docs, I still think there is a lot we could improve/change 🙂
All at it's time. For now dependency on ruby is gone.

forresst:
@jpsc I looked at the translation in French, it's ok for me

revelt:
Good stuff!

PR #902, opened 2016-10-31
Add python's list `append` method for Jinja compat

cflann:
## Summary

Adds python's [list.append](https://docs.python.org/2/library/array.html?highlight=append#array.array.append) for jinja compat.


vecmezoni:
Add a note to the changelog, please

cflann:
Updated.

vecmezoni:
Thank you!

Issue #912, opened 2016-11-13
Issues with passing variables of parent macro to child macro


ArmorDarks:
This issues seems to be related to changes of #667. Since Carl has been original author of PR and knows about Jinja2 much more than I do, I kindly ask @carljm to take a look at this issue.

As for now, in Nunjucks 3.0.0 there is no way to pass parameters of parent macro to its child macro.

Take an example:

``` jinja
{% macro Nav(argumentFromParent = 'Im argument from parent!') %}

    {% set setFromParent = 'Im set from parent' %}

    {% macro Item(argumentFromChildwithParentArgument = argumentFromParent, argumentFromChildwithParentSet = setFromParent) %}
      argumentFromParent                  : {{ argumentFromParent }}
      setFromParent                       : {{ setFromParent }}
      argumentFromChildwithParentArgument : {{ argumentFromChildwithParentArgument }}
      argumentFromChildwithParentSet      : {{ argumentFromChildwithParentSet }}
    {% endmacro %}

    {{ Item() }}

{% endmacro %}
```

As you can see, here I'm trying 4 ways of passing values from parrent macro (`Nav`), to it's child (`Item`):

1. Try read argument from parent directly (`argumentFromParent`)
2. Try read set in parent context variable directly (`setFromParent`)
3. Try to set argument from parent as default value of childs argument (`argumentFromChildwithParentArgument `)
4. Try to set set from parent context as default value of child argument (`argumentFromChildwithParentSet`)

All those attempts will print follwoing:

``` html
argumentFromParent                  : undefined
setFromParent                       : undefined
argumentFromChildwithParentArgument : undefined
argumentFromChildwithParentSet      : undefined
```

In other words, all attempts to pass variable from parent to child macro will fail.

It seems to me that 3 and 4 methods should work in Jinja2, and thus in Nunjucks too. Otherwise, is it indeed impossible to share values between macros in Jinja2?

@carljm Would you have time to take a look into Jinja2 behavior closer?

ArmorDarks:
Hm, this seems to be duplicate of https://github.com/mozilla/nunjucks/issues/906

But since my test case is slightly different, I will leave it open for now.

carljm:
This and #912 are clearly bugs. Don't know if they date from #667 or #791, but either way it looks like just unintended consequences of another fix, due to not having test cases for these cases. Jinja2 behavior is easy to check. Next step would be a PR with at the very least failing test cases.

ArmorDarks:
[...]

Issue #906, opened 2016-11-07
Block scoping is too strict (stricter than Jinja) in 3.0.0.


jgerigmeyer:
There were a number of commits included in 3.0.0 that fixed scoping issues, but now it seems that `{% call %}` blocks no longer have access to variables set in their parent scope. This differs from the Jinja2 implementation.

In Jinja2, the following code:

```django
{% macro inside() %}
  <div>inside: {{ caller() }}</div>
{% endmacro %}

{% macro outside(var) %}
  <div>outside: {{ var }}</div>
  {% call inside() %}
    {{ var }}
  {% endcall %}
{% endmacro %}

{{ outside('foobar') }}
```

Outputs:

```html
<div>outside: foobar</div>
<div>inside: foobar</div>
```

In Nunjucks 3.0.0, the output is:

```html
<div>outside: foobar</div>
<div>inside: </div>
```

See http://jsfiddle.net/jgerigmeyer/mjpbm13j/

ChristianAuth:
We got the same problem here. Not being able to access local variables inside of {% call %} blocks seems to break a lot of standard use cases. 

I would strongly vote to rollback the strict scoping for {% call %} blocks.

vecmezoni:
Introduced in https://github.com/mozilla/nunjucks/commit/429f636bde7a781b98849269560beb69388cba30

ArmorDarks:
Seems to be related to my similar issue https://github.com/mozilla/nunjucks/issues/912

paulmsmith:
This should be fixed but I'd love to [see the PR for adding `embed`](https://github.com/mozilla/nunjucks/pull/814) functionality progressed which would mean a lot of my use of `call` can be dropped entirely.

ArmorDarks:
`embed` isn't option for those, who works within multiple environments.

paulmsmith:
@ArmorDarks I get that, does that mean it's dead as a potential feature for Nunjucks? I think feature parity is important, could it be done as an extension perhaps?

ArmorDarks:
@paulmsmith Well, I'm not a maintainer, I can't tell :)

Personally I'm not very happy with `embed`, since it overlaps with macros. I think that instead macros should be enhanced. However, I understand needs of those, who goes from Swig to Nunjucks and needs `embed`, thus I don't vote against it.

I think for now it wasn't merged only because new maintainers didn't have enough time to review PR. My guess.

ArmorDarks:
Hi guys!

[...]

PR #917, opened 2016-11-18
Add failing test for passing vars from parent macro to child macro.

jgerigmeyer:
This PR simply adds a failing test for the most basic case of #912 (probably also related to #906).

### DIFF
--- a/tests/compiler.js
+++ b/tests/compiler.js
@@ -695,6 +695,20 @@
             finish(done);
         });
 
+        it('should allow child macro to access vars set in parent macro', function(done) {
+            equal('{% macro parent() %}' +
+                  '{% set x = "bar" %}' +
+                  '{% macro child() %}' +
+                  '{{ x }}' +
+                  '{% endmacro %}' +
+                  '{{ child() }}' +
+                  '{% endmacro %}' +
+                  '{{ parent() }}',
+                  'bar');
+
+            finish(done);
+        });
+
         it('should import templates', function(done) {
             equal('{% import "import.njk" as imp %}' +
                   '{{ imp.foo() }} {{ imp.bar }}',


ArmorDarks:
I'm not sure that all those cases are allowed in Jinja2, Some older commits, like https://github.com/mozilla/nunjucks/pull/667, clearly adding stronger isolation of child from parent scope. I guess to follow Jinja2 behavior better.

It's still unclear for me, does Jinja2 allows directly passing down variables to childs from parents, like in tests, or only via arguments, like in 3 and 4 test cases of #912.

Sorry @carljm for calling you once again, there isn't much Jinja2 experts around here :)

jgerigmeyer:
@ArmorDarks I tested all of the cases in your example #912; they all work as expected in Jinja2.

carljm:
Don't need to be a Jinja2 expert to test its behavior :-) Usually for these edge cases (I've personally never nested a macro definition inside another macro in real uses) I don't know the behavior, I have to check it. Here are the scripts I use for comparing Jinja2 behavior to nunjucks behavior on rendering some template:

```
$ cat jinja.py 
#!/usr/bin/env python
import sys
from jinja2 import Environment, FileSystemLoader

env = Environment(loader=FileSystemLoader('.'), autoescape=False)
tpl = env.get_template(sys.argv[1])
print (tpl.render({}))

[09:24] carljm@radicchio:~/p/n/test 
$ cat nunjucks.js 
#!/usr/bin/env node
var nunjucks = require('../nunjucks/index.js');

var env = nunjucks.configure('.');
var output = env.render(process.argv[2], {});
console.log(output);
```

The first script requires that you first create a Python virtualenv, activate it, and then run `pip install jinja2`. The second requires that Node is installed, and that a checkout of nunjucks is present in `../nunjucks/` (personally I have these two scripts and a bunch of test templates in a directory adjacent to my nunjucks checkout).

jgerigmeyer:
(I didn't add separate test cases for each example, but I can if you'd like.)

@carljm I agree that the examples in #912 are somewhat edge cases. #906, however, isn't at all. I assume they're related, but I could add a test specifically for #906 if you'd prefer that.

carljm:
I agree. They are quite probably related (IIRC effectively `{% call %}` defines a temporary un-named macro which is passed into the called macro as `caller`), although I think the codepaths are distinct enough from a regular macro definition that having a test specifically for that case is probably a good idea.

PR #930, opened 2016-12-18
Remove ancient Node.js compatibility workaround

shvaikalesh:


vecmezoni:
Could you please supply a note to the changelog?

shvaikalesh:
@vecmezoni I would gladly do that, but is necessary since `precompile` already uses `fs.existsSync` wo fallback: https://github.com/mozilla/nunjucks/blob/master/src/precompile.js#L42?

vecmezoni:
Ok, but i don't see any reason not to add a line to a changelog.

vecmezoni:
👍

PR #958, opened 2017-02-17
[WIP] Expose kwargs variable in macros context like jinja2

thibaudcolas:
## Summary

Proposed change:

This is a partial PR to discuss support for Jinja2's "special" macro variables `varargs` and `kwargs` (http://jinja.pocoo.org/docs/2.9/templates/#macros). Those variables, along with `caller` that Nunjucks supports already, are available in the context of a macro under specific conditions.

I'm unsure whether this is something Nunjucks wants to support (see #8), so at the moment this PR only adds very basic support for `kwargs`, adding the variable in the macro context but not fully conforming to the Jinja2 semantics. I'm keen to implement this more properly if there is agreement that this is valuable.

## Checklist

I haven't done any of this yet. For example the current implementation won't work in IE8 / Node 0.10 due to the use of `Object.assign`, so this will have to be changed.

~I've completed the checklist below to ensure I didn't forget anything. This makes reviewing this PR as easy as possible for the maintainers. And it gets this change released as soon as possible.~

### DIFF
--- a/src/compiler.js
+++ b/src/compiler.js
@@ -833,6 +833,12 @@ var Compiler = Object.extend({
             'var callerFrame = frame;',
             'frame = new runtime.Frame();',
             'kwargs = kwargs || {};',
+            // Expose kwargs variable in the context like Jinja2 macros.
+            'var kwargs_variable = Object.assign({}, kwargs);',
+            'delete kwargs_variable.__keywords;',
+            'delete kwargs_variable.caller;',
+            'frame.set("kwargs", kwargs_variable);',
+            // Expose caller variable in the context like Jinja2 macros.
             'if (kwargs.hasOwnProperty("caller")) {',
             'frame.set("caller", kwargs.caller); }'
         );


ArmorDarks:
Since times of #8 passed a lot of time, and, as far as I remember, @carljm and other were not against making Nunjucks close to Jinja2.

I'm voting for it, unless it makes performance worse.

thibaudcolas:
Thanks for the input! For performance, I'm not expecting this to have much of an impact either way. Looking at the [`makeMacro`](https://github.com/mozilla/nunjucks/blob/master/src/runtime.js#L87) code where this would be implemented, there is already some form of looping over the arguments given to the macro. The code of this feature will be very similar, and will only be executed for macros that take less arguments than they are given.

metatoaster:
@thibaudcolas are you also planning to basically support keyword expansion for macros (i.e. `macro(**kwargs)`)?

thibaudcolas:
@metatoaster is this value packing/unpacking (https://hangar.runway7.net/python/packing-unpacking-arguments) or something else? While it's a feature I use in my templates and I would like to see Nunjucks support it, no, I don't plan to introduce this as well as part of this PR.

As far as I understand this would be qualified as a (Python) language feature rather than a template feature, so this would warrant another conversation that I would rather keep out of this PR.

metatoaster:
@thibaudcolas Yes, that is it.  I saw the kwargs in the initial comment while searching for hints/progress/issues/pull requests related to that and I thought I might ask about the relation of this pull request to the value unpacking. Either way, thank you for your clarification on what this and isn't.

ArmorDarks:
Oh, I would like to see packing and unpacking in macros too so much! Though, not sure that they are supported in Jinja.

However, unfortunately, this issue isn't about it.

thibaudcolas:
This is now published as an npm package should anyone want to try it out:

```
npm install --save --save-exact @springload/nunjucks
```

maxnordlund:
This would be great, and I'm hoping this can be picked up again and merge :)

PR #976, opened 2017-03-31
Add python slice support to jinja-compat.

### LINKED ISSUE #188: Support Python's slice notation
It might be worth adding support for Python's slice notation (`foo[start:end:step]`)?

fdintino:
## Summary

Proposed change:

This pull-request adds support for python-style slice support (e.g. `arr[start:stop:step]`).

`jinja-compat.js` is the only file in this project that indents with two spaces, so my first commit changes it to four-space indent to bring it in conformity with the rest of the project.

There were not any existing tests for jinja compatibility, so I had to make some slight changes to the unit tests to have it work. Part of these changes was having `nunjucks.installJinjaCompat` return a function that, when called, undoes the jinja-compat changes, so that the jinja changes did not apply to other unit tests in the test suite. I wasn't sure if this was worth documenting in the API, since it strikes me as something that is unlikely to be used outside of the tests, but I would be happy to do so if requested.

Closes #188  .

## Checklist

With regard to the first bullet point above, support for python slice syntax is expressly listed as a feature which falls outside the purpose of Nunjucks. But the same developer that wrote that statement of purpose also suggested in #188 that, if this functionality were to be added, it should be added to jinja-compat. So I assumed that means that while it shouldn't be assumed that nunjucks would support language-specific syntax, it isn't forbidden either. Under this assumption I've added that clarification to the purpose section of CONTRIBUTING.md.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ master (unreleased)
 
 * Fix handling methods and attributes of static arrays, objects and primitives.
   Solves the issue [#937](https://github.com/mozilla/nunjucks/issues/937)
+* Add support for python-style array slices with Jinja compat enabled. Thanks
+  Frankie Dintino.
 
 3.0.0 (Nov 5 2016)
 ----------------
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -18,7 +18,7 @@ Nunjucks has the following purpose:
 
 Notes:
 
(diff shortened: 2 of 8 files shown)


fdintino:
@jbmoelker @vecmezoni wanted to make sure you saw this pull request. No rush of course.

zetareticoli:
When it'll be released?

Issue #972, opened 2017-03-22
Output nothing when using filter block in macro


ganlanyuan:
markup.njk:
```njk
{% macro get(blockName) %}

{% if blockName === 'html-indentation' %}
<ul>
..<li>Item</li>
..<li>Item</li>
</ul>

{% endif %}
{% endmacro %}
```

index.njk:
```njk
{% import 'code/markup.njk' as markup with context %}

{% macro outputCode(name) %}
  {# output nothing #}
  <pre><code>
  {%- filter trim %}{{ markup.get(name) }}{% endfilter -%}
  </code></pre>

  {# output content with 2 blank lines before and after the content #}
  <pre><code>
  {{- markup.get(name) + '&' | replace('&', '') -}}
  </code></pre>
{% endmacro %}

{% set str = "html-indentation" %}

{# output result #}
{{ outputCode('html-indentation') }}
```
When use a filter block inside the macro, nothing comes out. When remove the filter block, the content come out with 2 blank lines both before and after.

fdintino:
This is fixed by PR #990

ganlanyuan:
Thanks

PR #990, opened 2017-05-12
Addresses 914; capture blocks in macros now write to temporary output buffer

### LINKED ISSUE #914: Set blocks prints its content
In 3.0.0 set block for some reason immediately prints it content.

``` jinja2
{% set test %}
TEST
{% endset %}

result: {{ test }}
```

outputs

``` html
TEST

result:
```

As you can see, set print itself immediately, and `{{ test }}` call doesn't work.

Should it be like that?

noahlange:
## Summary

Proposed change: Fixes a bug that led the capture block compiler to write to the main buffer instead of the temporary one. Previously, the following code would (incorrectly) write "foobar".

```jinja
{%- macro foo(bar) -%}
  {%- set test -%}
    foo
  {%- endset -%}
  {{ bar }}{{ test }}
{%- endmacro -%}
{{ foo("bar") }}
```

Now it writes "barfoo."

Closes #914.

## Checklist

~I've completed the checklist below to ensure I didn't forget anything.~
( I accidentally hit enter before finishing the checklist.)

### DIFF
--- a/src/compiler.js
+++ b/src/compiler.js
@@ -1059,6 +1059,10 @@ var Compiler = Object.extend({
     },
 
     compileCapture: function(node, frame) {
+        // we need to temporarily override the current buffer id as 'output'
+        // so the set block writes to the capture output instead of the buffer
+        var buffer = this.buffer;
+        this.buffer = 'output';
         this.emitLine('(function() {');
         this.emitLine('var output = "";');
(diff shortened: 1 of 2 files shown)


noahlange:
This also appears to solve #972.

fdintino:
This looks great, thanks!

PR #1017, opened 2017-07-29
Support objects created with Object.create(null)

### LINKED ISSUE #468: obj.hasOwnProperty undefined is not a function nodejs
Hello,

I was noticing some discrepancies in some of the rendering of the nunjuck templates. After enough digging around I boiled it down to different nodejs versions between development and production. However the production machine has a more resent version of nodejs.

```
~ > node --version
v0.12.2
```

With version v0.12.x of node, this line

https://github.com/mozilla/nunjucks/blob/master/src/lib.js#L271

fails with **TypeError: undefined is not a function.**

Has anyone encountered this, is there any know workaround?

asprouse:
## Summary

Proposed change:

Support objects created with Object.create(null) by following the [no-prototype-builtins](http://eslint.org/docs/rules/no-prototype-builtins). This fix is important because it makes nunjucks compatible with this widely used ES5 feature (Object.create).

Closes #468

## Checklist

Note no documentation update was necessary.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ Changelog
 master (unreleased)
 -------------------
 
+* Support objects created with Object.create(null). fixes [#468](https://github.com/mozilla/nunjucks/issues/468)
+  
 
 3.0.1 (May 24 2017)
 -------------------
--- a/src/compiler.js
+++ b/src/compiler.js
@@ -838,7 +838,7 @@ var Compiler = Object.extend({
             'var callerFrame = frame;',
             'frame = ' + ((keepFrame) ? 'frame.push(true);' : 'new runtime.Frame();'),
             'kwargs = kwargs || {};',
-            'if (kwargs.hasOwnProperty("caller")) {',
+            'if (Object.prototype.hasOwnProperty.call(kwargs, "caller")) {',
             'frame.set("caller", kwargs.caller); }'
         );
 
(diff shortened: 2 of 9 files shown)


ArmorDarks:
This isn't related to PR directly, but I still can't get why would someone create object from `null`?

ArmorDarks:
Also, here https://github.com/mozilla/nunjucks/issues/468#issuecomment-115026511 said that this method doesn't work that well in older browsers.

asprouse:
@ArmorDarks creating an object using `Object.create(null)` creates an object without a prototype which is suitable for using as a map. Instead of calling `obj.hasOwnProperty(x)` to check membership one can simply `obj[x]`. It seems that this is a technique used is several popular libraries (express and graphql off the top of my head).

As far as compatibility goes this PR only changes how hasOwnProperty is called `obj.hasOwnProperty(x)` -> `Object.prototype.hasOwnProperty.call(obj, x)`. Both [hasOwnProperty](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwnProperty) and [call](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/call) predate ES5 (ES3 and ES1 according to MDN) so even IE8 support shouldn't be a problem.

fdintino:
Thanks for the PR! I agree that `Object.prototype.hasOwnProperty.call(obj)` is preferable to `obj.hasOwnProperty()`.

PR #1022, opened 2017-08-16
Add source for TemplateError in renderString

shybovycha:
## Summary

Proposed change: replace "unknown path" with source when rendering template string so when a rendering error occurs, it provides way more sense.

Compare the original error message to the new one:

```
Template render error: (unknown path) [Line 10, Column 76]
```

**vs**

```
Template render error: (Source: <p>Template text {{goes here}}</p>) [Line 10, Column 76]
```

### DIFF
--- a/src/environment.js
+++ b/src/environment.js
@@ -324,7 +324,7 @@ var Environment = Obj.extend({
         }
         opts = opts || {};
 
-        var tmpl = new Template(src, this, opts.path);
+        var tmpl = new Template(src, this, opts.path || ('Source:' + src));
         return tmpl.render(ctx, cb);
     },
 


ArmorDarks:
I think idea is quite cool, because so far debugging errors in templates (even not `renderString` used) is quite hard...

I think that source should be moved on newline, since in some cases it can be whole page, not just a single line of code.

And it would be so great if some sort of stack trace could be added... In jact, even native `new Error()` would be good, since it will contain information about JS file, which called `renderString`, thus making easier to get what exactly caused error.

shybovycha:
@ArmorDarks actually, after trying to debug an error generated by Hexo, I've noticed the errors generated are totally useless, since both `render` and `renderString` call the compiled template function. And making that debuggable is a way harder process. But I can look into it and make another PR =)

fdintino:
Thanks! Feel free to open a new PR with those improvements. In the meantime I'll close this pull request.

PR #1034, opened 2017-10-13
Adds switch tag, as featured in Craft CMS and Liquid. Addresses #967.

### LINKED ISSUE #967: Feature request: Add switch statement
Hello,

It would be great if there was a `{% switch %}` statement in nunjucks

Something like the following:

```
{% switch someVar %}
  {% case 'value1' %}
   ...
  {% case 'value2' %}
   ...
  {% case 'value3' %}
   ...
  {% default %}
   ...
{% endswitch %}
```

I'm relatively new to nunjucks so i'm not sure of the complexities involved in adding something like this.  I saw this: https://mozilla.github.io/nunjucks/api.html#custom-tags, but it's a little bit beyond me! 

Cheers

noahlange:
## Summary

Proposed change:

Adds a switch tag, as featured in [Liquid](http://shopify.github.io/liquid/tags/control-flow/#casewhen) and [Craft CMS](https://craftcms.com/docs/templating/switch), and as requested in [#967](https://github.com/mozilla/nunjucks/issues/967).

It works about how you'd expect. Switch on a variable, provide cases with potential values, and then an optional default case.

Still need to write docs + changelog entries.

**CAVEAT:** Unlike in JS, cases break automatically. There's no explicit break.
**CAVEAT**: Unlike in Liquid, switch tags use JS nomenclature (instead of Ruby's `case`/`when`/`else` ). It's much closer to Craft's implementation in that regard. If this is an issue, the code is written so that tag names can be easily swapped out.

```django
{% switch foo %}
  {% case 'bar' %}
    BAR
  {% case 'baz' %}
    BAZ
  {% default %}
    NEITHER BAR NOR BAZ
{% endswitch %}
```

The compiled code is fairly straightforward.

```js
function root(env, context, frame, runtime, cb) {
  var lineno = null;
  var colno = null;
  var output = '';
  try {
    var parentTemplate = null;
    switch (runtime.contextOrFrameLookup(context, frame, 'foo')) {
      case 'bar':
        output += 'BAR';
        break;
      case 'baz':
        output += 'BAZ';
        break;
      default:
        output += 'NEITHER FOO NOR BAR';
    }
    if (parentTemplate) {
      parentTemplate.rootRenderFunc(env, context, frame, runtime, cb);
    } else {
      cb(null, output);
    }
  } catch (e) {
    cb(runtime.handleError(e, lineno, colno));
  }
}

return {
  root: root
};
```
Closes #967 .

## Checklist

* [X] Proposed change helps towards [*purpose of this project*](https://github.com/mozilla/nunjucks/blob/master/CONTRIBUTING.md#purpose).
* [X] [*Tests*](https://github.com/mozilla/nunjucks/tree/master/tests) are added / updated to cover proposed change.

### DIFF
--- a/src/compiler.js
+++ b/src/compiler.js
@@ -565,7 +565,10 @@ var Compiler = Object.extend({
             this.compile(c.cond, frame);
             this.emit(': ');
             this.compile(c.body, frame);
-            this.emitLine('break;');
+            // preserve fall-throughs
+            if (c.body.children.length) {
+                this.emitLine('break;');
+            }
         }
         if (node.default) {
             this.emit('default:');
--- a/src/compiler.js
+++ b/src/compiler.js
@@ -555,6 +555,25 @@ var Compiler = Object.extend({
         }, this);
     },
 
+    compileSwitch: function(node, frame) {
+        this.emit('switch (');
+        this.compile(node.expr, frame);
+        this.emit(') {');
+        for (var i = 0; i < node.cases.length; i += 1) {
+            var c = node.cases[i];
+            this.emit('case ');
+            this.compile(c.cond, frame);
+            this.emit(': ');
+            this.compile(c.body, frame);
+            this.emitLine('break;');
+        }
+        if (node.default) {
+            this.emit('default:');
+            this.compile(node.default, frame);
+        }
+        this.emit('}');
+    },
+
     compileIf: function(node, frame, async) {
         this.emit('if(');
         this._compileExpression(node.cond, frame);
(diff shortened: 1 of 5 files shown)


ArmorDarks:
I'm somehow against going beyond what Jinja2 offers. Such kind of functionality should go through Jinja2 Github issues at first, where we can receive better community feedback, and only in case of approval implement it here.

Also, ignoring Jinja2 will take Nunjucks even further from original idea of compatible templating language. It would be understandable if Jinja2 is dead and does not evolve, thus holding Nunjucks back from further evolution. But it isn't dead.

Well, that's just imho.

noahlange:
Hmm. If it weren't a requested feature and implemented in two members of the Jinja family, I  wouldn't have bothered - I could probably rewrite it as a custom tag plugin.

fdintino:
I'm going to merge this as-is, with an eye towards making non-standard jinja tags into some sort of contrib module (either part of this repo, or as a separately installed repo).

revolter:
When will this be [documented](https://mozilla.github.io/nunjucks/templating.html)?

PR #1042, opened 2017-12-08
update cn templating.md

LongKissGoodBye:
there   should be `field`  instead of  `input`   in  "import"
 section


fdintino:
thanks!

Issue #1050, opened 2018-01-12
how to use with webpack?


TNT-Likely:

fdintino:
I'm afraid I don't understand your question. If you are looking for help setting up webpack, stack overflow might be a better resource.

PR #1052, opened 2018-01-17
Switch from jshint to eslint, enforce 2-space indent rules, fix linting errors

fdintino:
## Summary

Proposed change: switch from jshint to eslint (with an eye towards supporting es6 features in the future) and fix the inconsistency in indentation throughout the code base

In order to review substantive changes, append `?w=1` to the urls of the github diff view ([for example](https://github.com/mozilla/nunjucks/pull/1050/files?w=1))

This will likely cause merge conflicts with any currently open pull requests, so @ArmorDarks @noahlange (or anyone else) if there are any open pull requests that you think should be merged before this, please note this in the comments.

## Checklist

N/A


ArmorDarks:
I don't see any immediate PRs. In the worst scenario, it is quite trivial to update them to fit into new rules.

I'm definitely voting up for ESLint.

However, I'd recommend using [`standard`](https://standardjs.com/) instead, which is the same ESLint, but with tightened and commonly accepted rules, and many important additions. As an alternative, there is ESLint + Prettier, which is quite a comparable solution too.

In the worst case, I'd recommend using at least Airbnb ESLint rules, since they are much more popular than default recommended ESLInt rules.

What I definitely not recommending, is using own rules or overrides. There are already 3 popular standards — StandardJS, Airbnb and Prettier. There is no point in invention another one.

noahlange:
I'd strongly recommend ESLint + Prettier over Standard. AirBNB is really opinionated about not using ES2015+ features.

Standard is fine, but IMO, valuing code aesthetics over correctness makes it much less standard than it likes to think it is. For what it's worth, Prettier's also 3x+ more popular than Standard (according to State of JS 2017).

@fdintino - Thanks for doing the heavy lifting, by the way. It's a hell of a lot of code to sift through.

ArmorDarks:
> I'd strongly recommend ESLint + Prettier over Standard. AirBNB is really opinionated about not using ES2015+ features.

Yeap, regarding ES2015 features I agree. In such terms, Airbnb definitely not an option.

> Standard is fine, but IMO, valuing code aesthetics over correctness makes it much less standard than it likes to think it is

There is no such problem. At all.

Personally, I'm fine with both solutions, Standard or ESLint + Prettier. Both are nice. Both will make life better.

I slightly disliked Prettier because it still gives a configuration freedom, which kills the point of a standard. Rules aren't that tight. And integration with ESLint is a bit messy because it makes error reporting mostly useless. On contrary, with Prettier you would never read most part of the linting errors anyway.

> Thanks for doing the heavy lifting, by the way. It's a hell of a lot of code to sift through.

It would be so much easier to fix it automatically with `standard --fix` or Prettier formatting. Too late we've come to the party...

.eslintrc.js
```diff
@@ -0,0 +1,30 @@
...
+  'extends': 'eslint:recommended',
+    'parserOptions': {
+      'sourceType': 'module',
+      'ecmaVersion': 2017
+  },
+  'env': {
+    'node': true,
+    'es6': true,
+    'mocha': true,
+    'browser': true
```
ArmorDarks: I'd recommend not to write environment globally. 

You see, this can be an issue in files, which should, let's say, work only in browser env. ESLint will lint such as like they should work in Node too. And that might potentially leak some unexpected error. And current env will treat all files like test files, which imposes same risks. Though, I agree that it's very theoretical thing.

Not related to `es6`, of course.

fdintino:
> It would be so much easier to fix it automatically with standard --fix or Prettier formatting. Too late we've come to the party...

I definitely took advantage of `eslint --fix`. It's compatible with most of the rules I had in place.

I concur on using an established standard. After a bit of experimenting, I'm most happy with using both the standard config and `eslint-config-prettier`, which disables a bunch of pedantic rules in standard but retains the important ones (including some that I missed, such as `new-cap`, `no-unused-expressions`, `camelcase`, etc.).

.eslintrc.js
noahlange: ...could be worked into the build toolchain. we're already using webpack to bundle the web library, wouldn't be a huge jump to use it to bundle the server-side code as well. I'm mostly a TS guy, but if I recall correctly, eslint is pretty straightforward to integrate with Webpack.

ArmorDarks:
> including some that I missed, such as new-cap, no-unused-expressions, camelcase, etc.)

Just for the information, they are included in Standard too: https://github.com/standard/eslint-config-standard/blob/master/eslintrc.json

PR #1008, opened 2017-06-27
Update CONTRIBUTING.md

krassowski:
Fix broken links in contributing guideline.

## Summary

Proposed change: Fix broken links.

## Checklist

I've completed the checklist below to ensure I didn't forget anything. This makes reviewing this PR as easy as possible for the maintainers. And it gets this change released as soon as possible.

* [x] Proposed change helps towards [*purpose of this project*](https://github.com/mozilla/nunjucks/blob/master/CONTRIBUTING.md#purpose).
* [x] [*Documentation*](https://github.com/mozilla/nunjucks/tree/master/docs/) is added / updated to describe proposed change.
* [x] [*Tests*](https://github.com/mozilla/nunjucks/tree/master/tests) are added / updated to cover proposed change.
* [x] [*Changelog*](https://github.com/mozilla/nunjucks/blob/master/CHANGELOG.md) has an entry for proposed change (if user-facing fix or feature).

PR #1054, opened 2018-01-19
Fix a link in CONTRIBUTING.md

LoveIsGrief:
It was using the wrong markdown syntax.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -27,7 +27,7 @@ Issues and pull requests contributing to this purpose have the best chance to ma
 ## Questions?
 
 Please DO NOT ask "how do I?" or usage questions via GitHub issues. Instead,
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -27,7 +27,7 @@ Issues and pull requests contributing to this purpose have the best chance to ma
 ## Questions?
 
 Please DO NOT ask "how do I?" or usage questions via GitHub issues. Instead,
-use the [https://groups.google.com/forum/#!forum/nunjucks](mailing list).
+use the (mailing list)[https://groups.google.com/forum/#!forum/nunjucks].
```
ricordisamoa: Shouldn't parentheses and brackets be the other way around? – Already in #1008, by the way :sweat_smile:

LoveIsGrief: I'll let the other one be merged in then.

PR #1055, opened 2018-01-19
Differentiate variable blocks from simple output nodes

LoveIsGrief:
## Summary

Proposed change:

There doesn't seem to be any distinction between variable blocks `{{ }}`, blocks `{% %}` and other nodes text in the parser. In order to external actors to impose certain checks on the code that is going to be interpreted, they should be able to distinguish between Output nodes.

This pull request introduces the `VariableBlock` node. This will allow external actors to walk the AST and take custom actions on variable blocks.

For now, normal blocks aren't distinguished since the extension functionality seems to provide a way of overriding normal block handling, although I might be mistaken.

## Checklist

> Not sure about the changelog, since the parser and AST aren't fully documented despite their public interfaces.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ coverage
 .#*
 docs/_site
 docs/files
+.idea
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,3381 @@
+{
+  "name": "nunjucks",
+  "version": "3.0.2-dev.1",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "Base64": {
+      "version": "0.2.1",
+      "resolved": "https://registry.npmjs.org/Base64/-/Base64-0.2.1.tgz",
+      "integrity": "sha1-ujpCMHCOGGcFBl5mur3Uw1z2ACg=",
+      "dev": true
+    },
+    "a-sync-waterfall": {
+      "version": "1.0.0",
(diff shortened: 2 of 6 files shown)


package-lock.json
```diff
@@ -0,0 +1,3381 @@
+{
```
fdintino: While I agree that having a package-lock.json file is a good practice, I don't think it relates to the stated goal of the PR.

.gitignore
```diff
@@ -2,4 +2,5 @@ node_modules
 coverage
 .#*
 docs/_site
-docs/files
\ No newline at end of file
+docs/files
+.idea
```
fdintino: This too seems unrelated to the new Node types. There is a non-versioned gitignore file located at _.git/info/exclude_ if you ever have the need to ignore files that your text editor creates ([see here](https://help.github.com/articles/ignoring-files/), also, for how to set up global gitignores on your machine).

fdintino:
> For now, normal blocks aren't distinguished since the extension functionality seems to provide a way of overriding normal block handling, although I might be mistaken.

Could you detail a bit more how this is helpful? This specific implementation detail is patterned very closely on how Jinja2 does it [see here](https://github.com/pallets/jinja/blob/1f60059729fb4a49bfa32780844047b631941669/jinja2/nodes.py#L292-L296) and if anything I'd like to move this code towards _more_ fidelity to Jinja2, not less. That doesn't require that classes and objects are named the same thing in both codebases, but it does help.

PR #1063, opened 2018-02-09
Fix inappropriate term in "benefits" section

emceeaich:
The word "crazy" is a slur and not keeping with our community participation guidelines. https://www.mozilla.org/about/governance/policies/participation/

## Summary

Proposed change:

Closes # .

## Checklist

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -74,7 +74,7 @@ pageid: home
           control, and more. Heavily inspired by <a href="http://jinja.pocoo.org/">jinja2</a></li>
         <li><strong>Fast & Lean</strong> High-performant. Small 8K
           gzipped runtime with precompiled templates in the browser</li>
-        <li><strong>Extensible</strong> Crazy extensible with custom filters and extensions</li>
+        <li><strong>Extensible</strong> Write your own filters and extensions</li>
         <li><strong>Everywhere</strong> Available in node and all modern web browsers, with
           thorough precompilation options</li>
       </ul>


fdintino:
I will assume this pull request was opened in good faith, so I am happy to merge it. But it should be noted that "crazy" in the sense of "crazy extensible" is a north-americanism that simply means "extremely extensible" or "very extensible" and does not have even a remote relationship to mental illness. To be honest I would prefer changing it to say "Very extensible with with custom filters and extensions," both to retain the original meaning and also because it is a more similar construction to the other bullet points.

emceeaich:
I'm happy with "Very" as an intensifier here. Let me update the PR. Thanks!

Spown:
170 open tickets, 25 other prs, but no, this sjw nonsense is the only thing that matters.

I'm actually against this pr on the grounds of not caving in under pressure by people who aggressively seek to censor and thought police the Internet and the real life, by misusing the rules meant to be used in good faith. The word "crazy" is only then a slur when applied to a person and this particular case is not such an occasion. Don't get me wrong, the world would be a better place if no one would try to use derogative terms against other people, but this requires the elimination of this intention, not eradication of all "bad" words from the language (you can always invent more).

If the word "crazy" shifts into meaning "wild", "uncontrollable", "extensible", etc (which it was doing in the course of the last decades), slowly loosing it's applicability to living things, via organic process of language change - it will only be a good thing. When you try to interdict in this process with "cease & desist" you'll only make things worse. Not only will you remind about the word's insulting option (i.e. prevent it from fading), but you'll make it into an inciting taboo.

In short: no one was called crazy and nobody was trying to insult anybody. This is an appropriate usage of this word. A hundred thousand of such appropriate uses will make this word loose it's ability to be a slur. By inhibiting this process with censure you'll only achieve the opposite.

Please stop doing it.

fdintino:
@Spown I agree with some of what you wrote, but I'm still inclined to merge this pull request. I'd like to share my reasoning.

Given that this pull request exists, at least one person found the language objectionable. I think this is worth keeping in mind, and all other things being equal I think that's reason enough to merge. It doesn't cost me anything to do it. And, in fact, this change is an improvement for other reasons. A non-native English speaker reading these docs might not be familiar with this colloquial meaning of 'crazy.' In documentation, I think plain language is generally better.

That said, I myself have a serious mental illness, and I don't personally have an issue with the original language. I mention this because it's relevant here, but I have real anxiety about a current or future employer reading that admission. There is still a very real stigma around mental illness, and by comparison I tend to regard issues around language and usage as trivial. I think you're right that 'crazy extensible' here is pretty far removed from actual mental illness—it's intended to be read as an intensifier, akin to 'wildly,' and it actually carries a positive connotation here. But words can mean different things to different people.

If I might make a suggestion to @emceeaich, and I hope this isn't impertinent, I think that describing the wording as a 'slur' and 'inappropriate' isn't particularly helpful. I didn't write this documentation, but I can imagine that whoever did might take these characterizations personally—the implication being that the writer used an inappropriate slur, when I think it's safe to say that there was no such intention. To avoid knee-jerk defensiveness, I think it would be better to find more neutral ways of proposing changes such as this in the future.

fdintino:
@emceeaich can you check the box on this pull request to 'Allow edits from maintainers'? That way, I can resolve the conflict in the changelog for the merge.

PR #1072, opened 2018-02-21
Removed chokidar from optional dependencies

electerious:
`chokidar` is listed under dependencies and optional dependencies, but it's not optional (gets used in `node-loaders.js`). It's currently not possible to use Nunjucks with `npm --no-optional` as it will throw an `Cannot find module 'chokidar'` error.


fdintino:
Actually, I think this is a regression in 3.1.0. The require for chokidar should be inside a try/catch; I think that would be the appropriate fix.

PR #1126, opened 2018-07-12
Escape unicode newlines when compiling

### LINKED ISSUE #126: Some characters break Nunjucks
A U+2028 character (there may be others) in a template will cause nunjucks to throw "SyntaxError: Unexpected token ILLEGAL". Sometimes these sneaky beasts will work their way into copy. 

I understand if they cause an issue and throw or warn, but it should be handled better.

Tested on 0.1.10 and 0.1.9, broken in both.

EDIT: fixed character above; error happens regardless of where it is in the template.

### LINKED ISSUE #736: SyntaxError: Unexpected token ILLEGAL with unicode in HTML template
Not totally sure if this is a bug in Nunjucks or we're just using it wrong.

If the html template contains the [unicode character 8232](http://www.fileformat.info/info/unicode/char/2028/index.htm) (name: LINE SEPARATOR) then Nunjucks gives the error:

`SyntaxError: Unexpected token ILLEGAL`

TheDancingCode:
## Summary

Proposed change:

Escapes the `\u2028` unicode newline character to avoid errors when compiling.

Closes #126, closes #736 .

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 Changelog
 =========
 
+* Fix "Unexpected token" error for U+2028 unicode newline. Fixes [#126](https://github.com/mozilla/nunjucks/issues/126) and [#736](https://github.com/mozilla/nunjucks/issues/736)
+
 3.1.3 (May 19 2018)
 -------------------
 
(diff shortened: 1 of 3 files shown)


fdintino:
Thanks for this! Apologies for the delay in merging, it escaped my notice back in July.

fengmk2:
unicode newlines worked on node >= 10

Issue #1027, opened 2017-09-17
Please add "hello world" example to the docs


Ezra-Siton-UIX:
The site docs are really deep / well structure -  but it is very hard to find a simple/practical (html combine with template engine) example like (Template - data - output):
```
$ var name = 'Frank';
<h1>
  Hello ${name}!
</h1>
```
output: `<h1>Hello Frank!</h1>`

In nunjucks docs I find simple example with this sentence "`You usually won't use renderString`" ([Getting Started](https://mozilla.github.io/nunjucks/getting-started.html)) - but why you don't add example of what I **"usually should do"** :)  (codepen or any idea like this). This is really annoying. 

Examples for the first page **example** of docs (Simple and clear for new users **without** "won't")
- http://markojs.com/try-online/?file=%2Fmarko-language-guide%2Findex.marko&gist=
- https://codepen.io/SitePoint/pen/YyRxdP
- https://mustache.github.io/mustache.5.html
- http://www.dustjs.com/
- https://vuejs.org/v2/guide/#Declarative-Rendering
- https://blueimp.github.io/JavaScript-Templates/
- http://thunf.me/swiger/

adnion:
I very much support this idea! As a beginner I have a hard time getting a basic example up and running!

fdintino:
I'm a bit unclear about your criticism of the current docs. Here's the full passage quoted (emphasis added):

> You usually won't use renderString, **instead you should write templates in individual files and use render. That way you can inherit and include templates.** In this case, you need to tell nunjucks where these files live with the first argument of configure:
>
>     nunjucks.configure('views', { autoescape: true });
>     nunjucks.render('index.html', { foo: 'bar' });

That seems clear enough to me, but if you have a suggestion for how it could be improved I'd welcome a pull request.

yushiyangk:
How is that even remotely clear?

* Where do these two lines go?
* What is in index.html?
* What are in the templates?
* What is foo? What is bar?
* What is views?

It's not even clear from the documentation *how many files I should end up with*.

Your response to feedback that your documentation is not clear enough was basically "I understand it, so it's fine."

yushiyangk:
All you need to remedy this is to provide a *complete* hello world example, including each of the necessary files in their entirety, and instructions to build/execute it; no further explanation necessary.

fdintino:
[...]

PR #1187, opened 2019-02-01
add hello world example for issue #1027

devoidfury:
## Summary

Proposed change:

Add a simple hello world example to demonstrate basic usage in node with configure/render using the template directory.

Addresses #1027.


devoidfury:
I thought about trying to inline it, but thought it might get confusing with multiple files. Let me know if you have thoughts for improving this!

PR #1189, opened 2019-02-08
lexer more accurately tracks token line and column information

### LINKED ISSUE #1188: Token/Node lineno and colno do not correspond to source positions
I use nunjucks to power an internal tool at my work which allows users to configure nunjucks templates as part of a process automation tool. This has worked really well, and I've built a fair amount of tooling on top of nunjucks that uses the parser API to do some cool introspection to find e.g. references to symbols in the template which we know are invalid based on the context.

One frustration I've run into as I've used more internal APIs like the parser is that the Nodes' `lineno` and `colno` values don't seem to consistently match up with the source position in the input template. This has limited my ability to highlight error positions in the template, or at least required me to use some hacky workarounds to try to pinpoint a Node's source position.

After digging into it a bit, I think I've tracked the issue down to a single line in the lexer where the `index` field is updated directly in a way that does not keep `lineno` and `colno` in sync:

https://github.com/mozilla/nunjucks/blob/b82815815db5bb6249c270d518a77510fe575903/nunjucks/src/lexer.js#L378

This causes the line and column numbers to not be advanced to take into account tokens extracted with `_extractString` (primarily variable and block start/end tokens).

dylanscott:
## Summary

Proposed change:

This PR updates the lexer to more accurately keep track of tokens' source positions. The `_extractString` function was directly manipulating the `index` field without keeping `lineno` and `colno` in sync, whereas the rest of the lexer appears to use `forward` and `forwardN` which keeps line and column positions in sync.

Closes #1188 .

## Checklist

I'd be happy to update the documentation and/or changelog as necessary with this change. I'm not sure what the policy for internal API changes like this are.

### DIFF
--- a/nunjucks/src/lexer.js
+++ b/nunjucks/src/lexer.js
@@ -375,7 +375,7 @@ class Tokenizer {
 
   _extractString(str) {
     if (this._matches(str)) {
-      this.index += str.length;
+      this.forwardN(str.length);
       return str;
     }
     return null;
--- a/tests/compiler.js
+++ b/tests/compiler.js
@@ -816,7 +816,7 @@
         tmpl.render({}, function(err, res) {
           expect(res).to.be(undefined);
           expect(err.toString()).to.be([
-            'Template render error: (parse-error.njk) [Line 1, Column 24]',
+            'Template render error: (parse-error.njk) [Line 1, Column 26]',
             '  unexpected token: ,',
           ].join('\n'));
           done();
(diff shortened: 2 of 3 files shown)


fdintino:
This looks good to me. Not sure about the test failure, but it seems unrelated to your changes. I should probably remove node 6 from appveyor and travis anyway, since we no longer officially support it.

dylanscott:
Thanks! Let me know if there's anything I should do re: the build failure

fdintino:
Thank you for this!

PR #1196, opened 2019-03-04
Emit 'load' events on Loader and Environment instances

### LINKED ISSUE #1153: Provide a way to discover a template's dependencies
nunjucks doesn't provide a way to discover a template's direct or transitive dependencies i.e. the child and grandchild templates that are loaded (via `extends`, `import` or `include`) when a parent template is rendered. This is needed to [integrate](https://github.com/devmattrick/parcel-plugin-nunjucks/issues/1) ([efficiently](https://github.com/devmattrick/parcel-plugin-nunjucks/pull/3)) with bundlers such as [Parcel](https://github.com/parcel-bundler/parcel), which track changes in dependencies and use them to trigger rebuilds.

(Note: this is orthogonal to nunjucks' caching support via chokidar, though the implementations may overlap.)

Ideally, this would be scoped to a particular `Environment#render` call, but there doesn't appear to be any space in the `render` method's parameters to squeeze in additional options, or to return an additional value. Maybe a sister method with a more flexible interface?

```javascript
// result is an object rather than a string
const result = env.parse(templatePath, { context: ..., trackDependencies: true })
console.log('dependencies:', result.dependencies)
console.log('source:', result.source)
```

This makes dependency tracking optional, to err on the side of efficiency, but they could be included by default as it's a new API, and the overhead is unlikely to be high e.g.:

```javascript
const result = env.parse(templatePath, { context: ... })
```

Another way to do it is via an EventEmitter API (which is already used, lightly, in loaders) e.g.:

```javascript
env.on('dependency', ({ name, path, parent }) => { ... })
```

This has several advantages:

- it doesn't clutter or crowd any existing APIs
- it's easier to implement
- it's easy to extend
- if there's a performance concern, it could be toggled on/off with a constructor option e.g. `new Environment(loaders, { trackDependencies: true })`

The main disadvantage is that it doesn't *directly* answer the question "What are this template's dependencies?" i.e. it requires extra work to scope the results to a particular template/render. (This also implies the use of a full EventEmitter implementation rather than the cut-down version that's currently used for loaders, and suggests the addition of some new events e.g. `render:start` and `render:end`.)

---

I'm envisaging each dependency as an object with the following core fields:

```javascript
{
    name: "../macros/util.html.njk",
    path: "/home/foo/dev/example/src/html/macros/util.html.njk",
    parent: "/home/foo/dev/example/src/html/screens/layout.html.njk",
}
```

Optional fields could include `cached` (a boolean indicating that the template was retrieved from its loader's cache) and maybe `async` (the template was retrieved by an async loader).

# Example

### layout.html

```xml
{% include "../components/header.html" %}
<h1>Body</h1>
{% include "../components/footer.html" %}
```

### header.html

```xml
<h1>Header</h1>
```

### footer.html

```xml
<h1>Footer</h1>
{% include "./copyright.html" %}
```

### copyright.html

```xml
Copyright ⓒ example.com 2018
```

### render

```javascript
    const result = env.parse("src/html/screens/layout.html")
    console.log(result.dependencies)
```

### result

```javascript
[
    {
        name: "src/html/screens/layout.html",
        path: "/home/foo/example/src/html/screens/layout.html",
        parent: null,
    },
    {
        name: "../components/header.html",
        path: "/home/foo/example/src/html/components/header.html",
        parent: "/home/foo/example/src/html/screens/layout.html",
    },
    {
        name: "../components/footer.html",
        path: "/home/foo/example/src/html/components/footer.html",
        parent: "/home/foo/example/src/html/screens/layout.html",
    },
    {
        name: "./copyright.html",
        path: "/home/foo/example/src/html/components/copyright.html",
        parent: "/home/foo/example/src/html/components/footer.html",
    },
]
```

fdintino:
## Summary

Adds a 'load' event to `Environment` and `Loader` instances, to allow runtime dependency detection.

Closes #1153

## Checklist

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -353,6 +353,23 @@ SafeString (to be documented) if one was passed in, so the output will
 copy the safeness of the input, but this property is helpful in rare
 circumstances.
 {% endapi %}
+
+{% api %}
+'load' event
+env.on('load', function(name, source, loader))
+
+The 'load' event gets emitted whenever a Loader retrieves the source of a
+template. It can be listened to in order to determine template dependencies
+at runtime. The arguments emitted to the callback are:
+
+* **name** *(String)* The template name, as passed to the loader
+* **source** *(Object)* The object that gets returned from Loader.getSource
+  * **src** *(String)* The template source
+  * **path** *(String)* The full path to the template
+  * **noCache** *(Bool)* If `true`, the template wasn't cached.
+* **loader** The Loader instance that triggered the event.
+{% endapi %}
+
 {% raw %}
 
 ## Template
--- a/nunjucks/src/object.js
+++ b/nunjucks/src/object.js
@@ -1,6 +1,7 @@
 'use strict';
 
 // A simple class system, more documentation to come
+const EventEmitter = require('events');
 const lib = require('./lib');
 
 function parentWrap(parent, prop) {
@@ -59,4 +60,26 @@ class Obj {
   }
 }
 
-module.exports = Obj;
+class EmitterObj extends EventEmitter {
+  constructor(...args) {
+    super();
+    // Unfortunately necessary for backwards compatibility
+    this.init(...args);
+  }
+
+  init() {}
+
+  get typename() {
+    return this.constructor.name;
+  }
+
+  static extend(name, props) {
+    if (typeof name === 'object') {
+      props = name;
+      name = 'anonymous';
+    }
+    return extendClass(this, name, props);
+  }
+}
+
+module.exports = { Obj, EmitterObj };
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 Changelog
 =========
 
+* Emit 'load' events on `Environment` instances, to allow runtime dependency
+  tracking. Fixes [#1153](https://github.com/mozilla/nunjucks/issues/1153).
+
 3.1.7 (Jan 12 2019)
 ------------------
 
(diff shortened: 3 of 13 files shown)


docs/api.md
```diff
@@ -353,6 +353,23 @@ SafeString (to be documented) if one was passed in, so the output will
...
 circumstances.
 {% endapi %}
+
+{% api %}
+'load' event
+env.on('load', function(name, source, loader))
+
+The 'load' event gets emitted whenever a Loader retrieves the source of a
+template. It can be listened to in order to determine template dependencies
+at runtime. The arguments emitted to the callback are:
```
chocolateboy: s/emitted/passed/

nunjucks/src/object.js
```diff
@@ -59,4 +60,26 @@ class Obj {
   }
 }
 
-module.exports = Obj;
+class EmitterObj extends EventEmitter {
```
chocolateboy: Probably not worth changing now, but other EventEmitters are available, e.g. [little-emitter](https://www.npmjs.com/package/little-emitter), which is smaller and doesn't require the addition or modification of a superclass e.g. :

```js
const EventEmitter = require('little-emitter'); 

class Environment {
    constructor() {
        EventEmitter(this); // mix in `emit`, `on` etc.
    }
}
```

fdintino: I went with the default webpack events shim just because it's so widely used, it was easy, it's well documented, and it only added 1.5K gzipped to the minified builds.

PR #1197, opened 2019-03-05
Add NodeResolveLoader

### LINKED ISSUE #1175: [Feature Request] Resolver for NPM Modules
This has previously been raised as #1136 but is still an issue.

As previously raised:

> Suppose we have a package installed like this:
> 
> ```
> `npm i -S @example/cards`
> ```
> Now want to import from it like this within a template:
> 
> ```
> `{% import "@example/templates/birthday.html" as birthday %}`
> ```
> Could the nunjucks resolver implement NPMs [resolv alogrithm](https://nodejs.org/api/modules.html#modules_all_together)?

A potential solution has previously been suggested:
> `nunjucks.configure('node_modules');`
https://github.com/mozilla/nunjucks/issues/1136#issuecomment-421107441

However, as far as I know, this doesn't use standard node_modules resolution and so would not work under several conditions:

- using `npm link`
- having templates in peerDependencies, as the node_modules path could not be calculated relative to the current file
- where a project has module1 as a dependency, module1 uses nunjucks, which in turn has a dependency on module2 that contains templates. Depending on the dependency tree, the templates may live in project/node_modules/module2, or project/node_modules/module1/node_modules/module2
- when using [yarn pnp](https://github.com/yarnpkg/rfcs/pull/101), node_modules does not exist at a known location on disk
- when using [npm tink](https://blog.npmjs.org/post/178027064160/next-generation-package-management)
- in a monorepo (e.g. using lerna or yarn workspaces), which introduces several other patterns such as linking between modules

fdintino:
## Summary

Adds `NodeResolveLoader`, a Loader that loads templates using node's [`require.resolve`](https://nodejs.org/api/modules.html#modules_all_together).

Fixes #1175

## Checklist


fdintino:
The appveyor failure seems to indicate that I'm not using cross-env correctly to test this on windows. I'll need to investigate more later.

PR #1246, opened 2019-09-19
Throw existing render error instead of null

### LINKED ISSUE #1245: Error is lost on 'template not found' when rendering
When a 'template not found' error is generated by the render, the thrown error is lost and a `null` is thrown instead causing a `lineno of null` type error.

```
 TypeError: Cannot read property 'lineno' of null
    at Object.handleError node_modules/nunjucks/src/runtime.js:284:13)
    at Template.root [as rootRenderFunc] (eval at _compile (node_modules/nunjucks/src/environment.js:637:18), <anonymous>:175:39)
    at Template.getExported node_modules/nunjucks/src/environment.js:615:10)
    at eval (eval at _compile (node_modules/nunjucks/src/environment.js:637:18), <anonymous>:11:5)
    at Environment.getTemplate (node_modules/nunjucks/src/environment.js:279:9)
...
```

HughePaul:
## Summary

If there is an existing render error a flag is set so `cb()` isn't called twice, but when there is no `cb()` it tries to throw a new err, that might not exist and will therefore be null. This causes a `lineno of null` type error instead of seeing the render error message.

Proposed change:

Save off the previous error instead of just a boolean flag as to if an error occurred and throw that error.

Closes #1245 .

## Checklist

### DIFF
--- a/nunjucks/src/environment.js
+++ b/nunjucks/src/environment.js
@@ -473,20 +473,20 @@ class Template extends Obj {
     const frame = parentFrame ? parentFrame.push(true) : new Frame();
     frame.topLevel = true;
     let syncResult = null;
-    let didError = false;
+    let existingError = false;
 
     this.rootRenderFunc(this.env, context, frame, globalRuntime, (err, res) => {
-      if (didError) {
+      if (existingError) {
         // prevent multiple calls to cb
         if (cb) {
           return;
         } else {
-          throw err;
+          throw err || existingError;
         }
       }
       if (err) {
         err = lib._prettifyError(this.path, this.env.opts.dev, err);
-        didError = true;
+        existingError = err;
       }
 
       if (cb) {


fdintino:
Could you add a test case for this? (Or at least share in a comment an example of how you encountered this issue)?

fdintino:
I had tried to duplicate the problem myself, and found other bugs in how render errors surface, but not this particular one.

nunjucks/src/environment.js
```diff
@@ -473,20 +473,20 @@ class Template extends Obj {
...
-    let didError = false;
+    let existingError = false;
 
     this.rootRenderFunc(this.env, context, frame, globalRuntime, (err, res) => {
-      if (didError) {
+      if (existingError) {
         // prevent multiple calls to cb
         if (cb) {
           return;
         } else {
```
ogonkov: `else` is seems redundant

PR #1253, opened 2019-10-31
Replace yargs with commander to reduce dependencies

AlynxZhou:
## Summary

Replace [`yargs`](https://www.npmjs.com/package/yargs) with [`commander`](https://www.npmjs.com/package/commander), which is only used in `bin/precompile`. This PR does not break command interface, because they have the same logic and options.

Both yargs and commander can implement the same options for `bin/compile`, but [commander pulls in **0 dependency**](https://www.npmjs.com/package/commander?activeTab=dependencies) while [yargs pulls in **11 dependencies directly and maybe many others indirectly**](https://www.npmjs.com/package/yargs?activeTab=dependencies).

It seems most users of nunjucks use it as an library, not a cli-program, and they may not like many dependencies pulled in by yargs in their projects (like me).

## Checklist

* [X] Proposed change helps towards [*purpose of this project*](https://github.com/mozilla/nunjucks/blob/master/CONTRIBUTING.md#purpose).
* [X] [*Documentation*](https://github.com/mozilla/nunjucks/tree/master/docs/) is added / updated to describe proposed change.
* [X] [*Tests*](https://github.com/mozilla/nunjucks/tree/master/tests) are added / updated to cover proposed change.
* [X] [*Changelog*](https://github.com/mozilla/nunjucks/blob/master/CHANGELOG.md) has an entry for proposed change (if user-facing fix or feature).

### DIFF
--- a/bin/precompile
+++ b/bin/precompile
@@ -1,70 +1,52 @@
 #!/usr/bin/env node
-var path = require('path');
+var commander = require('commander');
 var precompile = require('../src/precompile').precompile;
 var Environment = require('../src/environment').Environment;
 var lib = require('../src/lib');
 
-var yargs = require('yargs')
+var cmdpath = null;
 
-    .usage('$0 [-f|--force] [-a|--filters <filters>] [-n|--name <name>] [-i|--include <regex>] [-x|--exclude <regex>] [-w|--wrapper <wrapper>] <path>')
-    .wrap(80)
+commander
+  .name('precompile')
+  .usage('[-f|--force] [-a|--filters <filters>] [-n|--name <name>] [-i|--include <regex>] [-x|--exclude <regex>] [-w|--wrapper <wrapper>] <path>')
+  .arguments('<path>')
+  .helpOption('-?, -h, --help', 'Display this help message')
+  .option('-f, --force', 'Force compilation to continue on error')
+  .option('-a, --filters <filters>', 'Give the compiler a comma-delimited list of asynchronous filters, required for correctly generating code')
+  .option('-n, --name <name>', 'Specify the template name when compiling a single file')
+  .option('-i, --include <regex>', 'Include a file or folder which match the regex but would otherwise be excluded. You can use this flag multiple times', concat, ['\\.html$', '\\.jinja$'])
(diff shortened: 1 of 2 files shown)


bin/precompile
```diff
@@ -1,70 +1,50 @@
...
-
-var argv = yargs.argv;
+function concat(value, previous) {
+  return previous.concat(value);
+}
 
-if (argv.help) {
-    yargs.showHelp();
-    process.exit(1);
+commander
```
ogonkov: it would be better to move `concat` function after this `commander` invocation, to keep diff less complicated

AlynxZhou: Some `.option()` call needs `concat()` as an argument, maybe it's not so good to move it after calling, I think.

ogonkov: In JavaScript functions get hoisted, so it should work

nikku:
If you really want to go lightweight, consider going for [mri](https://packagephobia.now.sh/result?p=mri) instead of [commander](https://packagephobia.now.sh/result?p=commander).

AlynxZhou:
No, I am not talking about lightweight, commander has 0 dependency while
yargs has a lot, that's enough.

Also I noticed that last commit of mri happens one year ago, so maybe it's
not good to use it instead of commander now.

I am not familiar with mri, so I won't replace commander with it. Sorry.

nikku:
[`mri`](https://www.npmjs.com/package/mri) is a drop-in replacement for `yargs` in most cases and it is zero dependency, too.

AlynxZhou:
> [`mri`](https://www.npmjs.com/package/mri) is a drop-in replacement for `yargs` in most cases and it is zero dependency, too.

Thank you for your advice. I'm just not familiar with mri, and I am familiar with commander, so I write patch with what I can do better. I'm just not sure whether I can make patch with library that I never use.

Last commit of mri happened in Dec 24, 2018, it makes people unsure whether it's still maintaining. Also when we refer to 0 dependency command-line arguments parser, commander has been used widely, but mri is not compared with commander. That's the reason why I choose commander.

bin/precompile
AlynxZhou: I've moved it. Thank you for your advice.

PR #1276, opened 2020-03-29
Add support for nested attributes in `groupBy` filter

### LINKED ISSUE #1198: groupby complex attributes
I have a list of objects. These objects have a `date` attribute. I want to group by year.
Something along the lines of the following would be nice.

```
for year, posts in collections.post | groupby("date.year")
```

But it begs the question how to deal with transforms that need to happen before the grouping.
Is there a way to use `groupby` like this?

ogonkov:
## Summary

Proposed change:

Add support for nested attributes access for `groupBy` filter to match [Jinja2 behaviour](https://jinja.palletsprojects.com/en/2.11.x/templates/#groupby)

Closes #1198 .

## Checklist

### DIFF
--- a/nunjucks/src/lib.js
+++ b/nunjucks/src/lib.js
@@ -167,12 +167,58 @@ function isObject(obj) {
 
 exports.isObject = isObject;
 
-function groupBy(obj, val) {
+/**
+ * @param {string|number} attr
+ * @returns {(string|number)[]}
+ * @private
+ */
+function _prepareAttributeParts(attr) {
+  if (!attr) {
+    return [];
+  }
+
+  if (typeof attr === 'string') {
+    return attr.split('.');
+  }
+
+  return [attr];
+}
+
+/**
+ * @param {string}   attribute      Attribute value. Dots allowed.
+ * @returns {function(Object): *}
+ */
+function getAttrGetter(attribute) {
+  const parts = _prepareAttributeParts(attribute);
+
+  return function attrGetter(item) {
+    let _item = item;
+
+    for (let i = 0; i < parts.length; i++) {
+      const part = parts[i];
+
+      // If item is not an object, and we still got parts to handle, it means
+      // that something goes wrong. Just roll out to undefined in that case.
+      if (hasOwnProp(_item, part)) {
+        _item = _item[part];
+      } else {
+        return undefined;
+      }
+    }
+
+    return _item;
+  };
+}
+
+function groupBy(obj, val, throwOnUndefined) {
   const result = {};
-  const iterator = isFunction(val) ? val : (o) => o[val];
+  const iterator = isFunction(val) ? val : getAttrGetter(val);
   for (let i = 0; i < obj.length; i++) {
     const value = obj[i];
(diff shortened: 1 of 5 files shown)


ogonkov:
@fdintino should i make PR against your dev branch aswell?

anaurelian:
This is the code I'm using to achieve the same effect:

This is a filter I'm using for Nunjucks in [Eleventy](https://www.11ty.dev/):

    groupByEx: function (arr, key) {
        const result = {};
        arr.forEach(item => {
            const keys = key.split('.');
            const value = keys.reduce((object, key) => object[key], item);

            (result[value] || (result[value] = [])).push(item);
        });
        return result;
    }

Sample (actual) usage:

`for categoryName, articles in productArticles | groupByEx("data.categoryName")`

Acknowledgement: Code based on [whatterz/includes.js](https://gist.github.com/whatterz/df2817ba36e57617a17123c709fff792)

ogonkov:
Now it should work out of the box (when it gets merged)

fdintino:
Don't worry about the conflicts, I'll take care of those. I'll try to set aside some time, if not today, then this week to give feedback or merge.

ogonkov:
@fdintino can you start from `reject` PR, please? )

fdintino:
Sure, will do.

ogonkov:
@fdintino this PR is ready for review ;)

nunjucks/src/lib.js
```diff
@@ -167,9 +167,58 @@ function isObject(obj) {
...
+
+    for (let i = 0; i < parts.length; i++) {
+      const part = parts[i];
+
+      // If item is not an object, and we still got parts to handle, it means
+      // that something goes wrong. Just roll out to undefined in that case.
+      if (hasOwnProp(_item, part)) {
+        _item = _item[part];
+      } else {
+        rollback();
```
fdintino, line 209: What's the advantage of the rollback function? Wouldn't it be functionally equivalent to simply `return;` here?

fdintino:
A small observation: the behavior here doesn't respect `opts.throwOnUndefined`. The previous implementation didn't either, so I don't think you need to address that in this PR. It does mean that it behaves differently from jinja2, where attempts to groupby a non-existent attribute raises an exception. It's regrettable that nunjucks doesn't have its own `Undefined` object, distinct from 'undefined', the way that jinja2 does.

If you did want to use the value of `Environment.opts.throwOnDefined`, you would be able to get it from `this.env.throwOnUndefined` in the filter function.

nunjucks/src/lib.js
ogonkov: You right, It was added in previous version, when there was default value in arguments. I will replace it with return for now.

ogonkov:
@fdintino i have add support for this option. I just throw it inside function, hope it would be handled somewhere above.

ogonkov:
@fdintino everything is OK?

fdintino:
Yep, everything is fine. Sorry, had to step away for a bit when I was in the middle of preparing for merge.

PR #1278, opened 2020-03-29
Add `select` filter

ogonkov:
## Summary

Proposed change:

* Add [`select` filter](https://jinja.palletsprojects.com/en/2.11.x/templates/#select)

Closes #282 .

## Checklist

I've completed the checklist below to ensure I didn't forget anything. This makes reviewing this PR as easy as possible for the maintainers. And it gets this change released as soon as possible.

* [x] Proposed change helps towards [*purpose of this project*](https://github.com/mozilla/nunjucks/blob/master/CONTRIBUTING.md#purpose).
* [x] [*Documentation*](https://github.com/mozilla/nunjucks/tree/master/docs/) is added / updated to describe proposed change.
* [x] [*Tests*](https://github.com/mozilla/nunjucks/tree/master/tests) are added / updated to cover proposed change.
* [x] [*Changelog*](https://github.com/mozilla/nunjucks/blob/master/CHANGELOG.md) has an entry for proposed change (if user-facing fix or feature).

PR #1279, opened 2020-03-29
Add `reject` filter

### LINKED ISSUE #282: Implement select/reject
These exists in jinja and is a pretty good way to exclude something from array iteration. I'm working on a patch.

ogonkov:
## Summary

Proposed change:

Add [`reject` filter](https://jinja.palletsprojects.com/en/2.11.x/templates/#reject) to match Jinja2 behaviour.

Closes #282.

## Checklist

### DIFF
--- a/nunjucks/src/filters.js
+++ b/nunjucks/src/filters.js
@@ -266,22 +266,34 @@ function random(arr) {
 
 exports.random = random;
 
+/**
+ * Construct select or reject filter
+ *
+ * @param {boolean} expectedTestResult
+ * @returns {function(array, string, *): array}
+ */
+function getSelectOrReject(expectedTestResult) {
+  function filter(arr, testName = 'truthy', secondArg) {
+    const context = this;
+    const test = context.env.getTest(testName);
+
+    return lib.toArray(arr).filter(function examineTestResult(item) {
+      return test.call(context, item, secondArg) === expectedTestResult;
+    });
+  }
+
+  return filter;
+}
+
+exports.reject = getSelectOrReject(false);
+
 function rejectattr(arr, attr) {
   return arr.filter((item) => !item[attr]);
 }
 
 exports.rejectattr = rejectattr;
 
-function select(arr, testName = 'truthy', secondArg) {
-  const context = this;
-  const test = context.env.getTest(testName);
-
-  return arr.filter(function applyToTest(item) {
-    return test.call(context, item, secondArg);
-  });
-}
-
-exports.select = select;
+exports.select = getSelectOrReject(true);
 
 function selectattr(arr, attr) {
   return arr.filter((item) => !!item[attr]);
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,16 @@
 Changelog
 =========
 
-* Add [`select` filter](https://mozilla.github.io/nunjucks/templating.html#select).
-* Add [`reject` filter](https://mozilla.github.io/nunjucks/templating.html#reject).
+3.2.2 (unreleased)
+------------------
+
+* Add [`select`](https://mozilla.github.io/nunjucks/templating.html#select) and
+  [`reject`](https://mozilla.github.io/nunjucks/templating.html#reject) filters.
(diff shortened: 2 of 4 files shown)


ogonkov:
It's better to merge it after #1278 since they share many common code.

ogonkov:
@fdintino now it's ready.

Codecov shows uncovered lines for file that is not in my PR for some reason

ogonkov:
@fdintino Can you please trigger build in Travis for master it should fix coverage

ogonkov:
@fdintino ?

nunjucks/src/filters.js
```diff
@@ -266,20 +266,34 @@ function random(arr) {
...
+      return testResultFn(test.call(context, item, secondArg));
+    });
+  }
+
+  return filter;
+}
+
+const reject = getSelectOrReject(function examineRejectResult(testResult) {
+  return testResult === false;
+});
```
fdintino, line 284: It looks like jinja2 doesn't strictly require reject to return false from the test function, it's sufficient for it to be truthy. So it might be better to do this as:

```suggestion
const reject = getSelectOrReject((res) => !res);
const select = getSelectOrReject((res) => !!res);
```

ogonkov: `testResult` here is the output of `test` function, so it is always boolean.

ogonkov: I have add jsdoc to make it more clear

fdintino: Thanks!

nunjucks/src/filters.js
```diff
@@ -266,20 +266,40 @@ function random(arr) {
...
+ *
+ * @param {function(boolean): boolean} testResultFn
+ * @returns {function(array, string, *): array}
+ */
+function getSelectOrReject(testResultFn) {
+  function filter(arr, testName = 'truthy', secondArg) {
+    const context = this;
+    const test = context.env.getTest(testName);
+
+    return arr.filter(function applyToTest(item) {
```
fdintino, line 280: This can be any iterable, not necessarily just an array, right? For example, this template:

```jinja2
{{ vals | reject("sameas", "0") | join("") }}
```

with `vals = '012301230123'`

ogonkov: Looks like you right. It not clear from Jinja2 docs, but they have code example that used `for..in`, that seems to work for strings aswel.

I will change to `for` loop.

ogonkov: But in Jinja2 test they are testing only [against lists](https://github.com/pallets/jinja/blob/6478c22f29bb85b2bae635602148d95df5e5c7ce/tests/test_filters.py#L664-L672).

ogonkov: Can we keep only simple case with arrays?

Adding support for strings `select`/`reject` could be next step, if community will request it.

fdintino: I think you could do this rather simply by replacing `arr` with `Array.from(arr)`

fdintino: Something like this would do the trick:

```javascript
function getSelectOrReject(isTrue) {
  return function filter(arr, testName = 'truthy', ...args) {
    const test = this.env.getTest(testName);
    return Array.from(arr).filter(item =>
      (isTrue === test.apply(this, [item].concat(args))));      
  };
}

exports.select = getSelectOrReject(true);
exports.reject = getSelectOrReject(false);
```

ogonkov: `Array.from` is failed in Phantom. Is there any way to inject polyfills to tests bundle?

fdintino: Oh, right. Try importing and using `toArray` from `lib.js`

ogonkov: Done

ogonkov:
@fdintino can we finish this Pull Request? :)

fdintino:
@ogonkov : absolutely! My apologies though: it may not be for a week or two, at least. Only just got back to work from paternity leave, still managing remote work with a newborn and getting through a backlog of open source project maintenance.

PR #1280, opened 2020-03-29
Decrease NPM package size

### LINKED ISSUE #984: Reduce package size
Package contains unnecessary directories and files - `.github`, `.idea` and etc. I can do PR if your accept this issue.

ogonkov:
## Summary

Proposed change:

Strip development files from NPM package.

Closes #984.

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,7 +79,12 @@
   "bin": {
     "nunjucks-precompile": "./bin/precompile"
   },
-  "main": "index",
+  "main": "index.js",
+  "files": [
+    "bin/**",
+    "browser/**",
+    "src/**"
+  ],
   "nyc": {
     "require": [
       "babel-register"


ogonkov:
@fdintino `samples` dir is likely should be ignored aswell?

.npmrc
```diff
@@ -0,0 +1 @@
+package-lock=false
```
mroderick, line 1: What is the motivation? Why would you not want a `package-lock.json`?

In other projects, we're very happy to have repeatable builds thanks to having a `package-lock.json` artifact in the repository.

package.json
```diff
@@ -87,6 +87,7 @@
     "sourceMap": false,
     "instrument": false
   },
+  "homepage": "https://mozilla.github.io/nunjucks/",
```
mroderick, line 90: This could go into a separate PR, that might get merged faster

.npmrc
ogonkov: Locking dependencies is good for apps, but not suitable for libraries, since it's not distributed with package. End-users and developer will end with different dependency tree, that could lead to bugs.

I have nice example. With one of the latest PR i got `mocha` failed to run tests, while it works perfectly on my local machine. After removing `package-lock.json` i was able to reproduce problem.

package.json
ogonkov: Probably, but maintainer got no time at that moment, he promise to sort out PRs later.

.npmrc
fdintino: Actually, I think best practices today are that you commit the package-lock.json (or yarn.lock, if using yarn) to the repository. You exclude it via .npmignore, and even if nunjucks is installed as a peer dependency directly from git, it won't look to the package-lock.json file. It is only used for development. So, if the package-lock.json file had been committed to git, you wouldn't have encountered the mocha issue.

ogonkov: When one of the `dependencies` will get new version, you didn't get it locally, like package user do.

Even CI would not catch a bug, because it will resolve version from lock file, while package users will resolve new dependency version from `package.json`.

Lock file is bad idea for libraries. It good for apps, that should remain same.

fdintino: That is only true of the packages in `dependencies`, of which there are very few. By including lock files, we would be following the lead of the largest node projects: vue.js, vscode, React, Nest, Angular, next.js, and many more—all of whom have either a yarn.lock or a package-lock.json file committed to the repository.

Typically, adding a lock file to source control is done alongside integrating with something like dependabot. These bots open separate pull requests for every dependency update. This allows you to identify and fix bugs caused by upgrades (because they cause CI failures) _before_ they start showing up in unrelated pull requests of contributors.

mroderick: We use lock files in the Sinon projects, in order to have repeatable builds.

package.json
mroderick: If it's a one-liner, then it's much more likely to get merged without delay

.npmrc
ogonkov: I have remove this file. But i still believe that `package-lock.json` is redundant for a library. But i personally like what this guy wrote https://github.com/sindresorhus/ama/issues/479#issuecomment-310661514

package.json
ogonkov: Moved to #1299

rdmurphy:
This works, but couldn't we do the inverse and set the [`files` field in the `package.json`](https://docs.npmjs.com/files/package.json#files) so new additions to the repo don't require being kept in sync with the `.npmignore`? I bet that'd be easier to keep up to date.

ogonkov:
Good idea

ogonkov:
Package after adding `files` field
<details>
<pre>
npm notice 
npm notice 📦  nunjucks@3.2.1
npm notice === Tarball Contents === 
npm notice 1.3kB   LICENSE                         
npm notice 2.1kB   bin/precompile                  
npm notice 111B    bin/precompile.cmd              
npm notice 37.9kB  src/compiler.js                 
npm notice 16.5kB  src/environment.js              
npm notice 705B    src/express-app.js              
npm notice 14.6kB  src/filters.js                  
npm notice 1.4kB   src/globals.js                  
npm notice 2.3kB   index.js                        
npm notice 8.5kB   src/jinja-compat.js             
npm notice 15.8kB  src/lexer.js                    
npm notice 7.9kB   src/lib.js                      
npm notice 783B    src/loader.js                   
npm notice 151B    src/loaders.js                  
npm notice 4.5kB   src/node-loaders.js             
npm notice 9.7kB   src/nodes.js                    
npm notice 103.9kB browser/nunjucks-slim.js        
npm notice 35.5kB  browser/nunjucks-slim.min.js    
npm notice 216.0kB browser/nunjucks.js             
npm notice 88.0kB  browser/nunjucks.min.js         
npm notice 3.2kB   src/object.js                   
npm notice 35.4kB  src/parser.js                   
npm notice 602B    src/precompile-global.js        
npm notice 3.7kB   src/precompile.js               
npm notice 906B    src/precompiled-loader.js       
npm notice 9.0kB   src/runtime.js                  
npm notice 5.8kB   src/tests.js                    
npm notice 5.2kB   src/transformer.js              
npm notice 3.0kB   src/web-loaders.js              
npm notice 3.1kB   package.json                    
npm notice 185.7kB browser/nunjucks-slim.js.map    
npm notice 151.5kB browser/nunjucks-slim.min.js.map
npm notice 406.2kB browser/nunjucks.js.map         
npm notice 326.2kB browser/nunjucks.min.js.map     
npm notice 28.7kB  CHANGELOG.md                    
npm notice 2.0kB   README.md                       
npm notice === Tarball Details === 
npm notice name:          nunjucks                                
npm notice version:       3.2.1                                   
npm notice filename:      nunjucks-3.2.1.tgz                      
npm notice package size:  480.4 kB                                
npm notice unpacked size: 1.7 MB                                  
npm notice shasum:        caea4f0663e0ae567409a0b126bfb4370d463542
npm notice integrity:     sha512-4PADUSdpug/XO[...]DHp0CF4tP//Iw==
npm notice total files:   36                                      
npm notice 
nunjucks-3.2.1.tgz
</pre>
</details>

fdintino:
Thank you! I'll merge this into master shortly.

package.json
```diff
@@ -79,7 +79,12 @@
   "bin": {
     "nunjucks-precompile": "./bin/precompile"
   },
-  "main": "index",
+  "main": "index.js",
+  "files": [
+    "bin/**",
+    "browser/**",
+    "src/**"
```
fdintino, line 86: This also removes the files in `nunjucks/src` from the package. In theory that should be safe, but it would be a breaking change (if somehow, somebody was importing those files). To err on the side of caution we should probably continue to ship them in the package, at least until the next minor version release.

ogonkov: That was intended change. This files was never officially provided by the package, if someone hack into package content, he definitely know about risks. I think we could safely make this change, this will save all other users a lot.

ogonkov: Also, same files in `src` dir

fdintino: They are not exactly the same files, because they're transpiled, but you make a decent point. If we get a bunch of error reports we can always put out another patch release that adds it back in.

PR #1296, opened 2020-07-07
Fix #1295

### LINKED ISSUE #1295: precompile.js TypeError: name.replace is not a function
I'm getting a runtime error running `precompile.js` on line 114.

Command line:

```
$ node_modules/nunjucks/bin/precompile input_template.xml > output_template.js
```

gives a TypeError in the function `_precompile` when it tries to call:

```js
  name = name.replace(/\\/g, '/');
```

Debugging, this is because further up the stack, in the `precompile` function, line 85, the call to `_precompile` passes in as the second argument `opts.name || input`. Whilst `input` is indeed my input filename string, `opts.name` is a function pointer, not a string. Replacing this text with `opts.name() || input` appears to fix the problem, though I'm not familiar enough with the codebase to understand if that is the right place to make the fix.

In my case `opts.name()` returns `undefined`, which means the input filename gets used instead.

nunjucks release 3.2.1
node v12.7.0

Stack trace:

```
  name = name.replace(/\\/g, '/');
              ^

TypeError: name.replace is not a function
    at _precompile (/XXX/node_modules/nunjucks/src/precompile.js:114:15)
    at precompile (/XXX/node_modules/nunjucks/src/precompile.js:85:22)
    at Object.<anonymous> (/XXX/node_modules/nunjucks/bin/precompile:45:13)
    at Module._compile (internal/modules/cjs/loader.js:777:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:788:10)
    at Module.load (internal/modules/cjs/loader.js:643:32)
    at Function.Module._load (internal/modules/cjs/loader.js:556:12)
    at Function.Module.runMain (internal/modules/cjs/loader.js:840:10)
    at internal/main/run_main_module.js:17:11
```

nigelmegitt:
## Summary

Proposed change:

In `precompile` when we're using a file, call `opts.name()` before testing if the result is undefined rather than passing the function pointer downstream. This fixes a bug where `name.replace` was failing because it was being called on a function pointer rather than a string.

Closes #1295 .

Note that an alternative solution to this would be to make `opts.name` a getter; I did not investigate that option because I wanted to make the smallest change that would fix this bug.

## Checklist

* [x] [*Documentation*](https://github.com/mozilla/nunjucks/tree/master/docs/) is added / updated to describe proposed change. **No documentation to update; this is a bug fix only**
* [x] [*Tests*](https://github.com/mozilla/nunjucks/tree/master/tests) are added / updated to cover proposed change. **It looks like there are already tests, I'm not sure why they don't cover this issue**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 Changelog
 =========
 
+* Fix bug where precompiling a template from a file failed with a TypeError. [#1295](https://github.com/mozilla/nunjucks/issues/1295)
 * Add [`select` filter](https://mozilla.github.io/nunjucks/templating.html#select).
 
 3.2.1 (Mar 17 2020)
--- a/nunjucks/src/precompile.js
+++ b/nunjucks/src/precompile.js
@@ -73,7 +73,7 @@ function precompile(input, opts) {
   if (pathStats.isFile()) {
     precompiled.push(_precompile(
       fs.readFileSync(input, 'utf-8'),
-      opts.name || input,
+      opts.name() || input,
       env
     ));
   } else if (pathStats.isDirectory()) {


nigelmegitt:
This fix causes existing tests to fail. The correct solution to fix #1295 may be more subtle: whatever is making `opts.name` a function pointer is probably the source of this bug, not its usage as is.

PR #1299, opened 2020-07-17
Add `homepage` to `package.json`

ogonkov:
## Summary

Proposed change:

Improve package a bit, by adding link to actual home page of the package.

## Checklist


mroderick:
Very nice 👍

fdintino:
Running the build in prepare is intentional. We don't commit the build files to version control. So if we didn't have the prepare script in package.json, it would not be possible to install nunjucks directly from github (which there are valid reasons for people to want to be able to do). Are you seeing the build scripts running when you do an `npm install nunjucks` (in other words, when installing from the npm registry)? Because I'm not seeing that.

ogonkov:
I was not aware of that feature, with installing from git. I see build for every local `npm install`, when installing dependencies for development, that annoys me.

I think installing from git is nice feature. By looking to [docs](https://docs.npmjs.com/misc/scripts), it seems that we could safely change `prepare` to `prepack`, to keep it building on git install, but not on local `npm install`s.

Krinkle:
If you only see it when running `npm install` on the nunjucks directory (e.g. when working on it as a contributor, not when installing it as dependency elsewhere), then that seems useful. The only use case for doing that afaik is when working on the code itself, at which point anything you do with it, wil require a build. I think it would be confusing to contributors and maintainers if `npm installl && npm test` fails. The build step should either happen via `prepare` as currently done (if it's needed for releases and github installs), or as part of `test` if it is only needed during development. Running `build` manually would imho be unexpected.

If it is running in cases where it is not needed, perhaps we can find a way to avoid it in that specific circumstance only.

PR #1301, opened 2020-07-18
Fix issue where sync render would not raise errors in included templates

### LINKED ISSUE #1272: Exceptions being swallowed after import errors in included templates
As stated in topic #1127, there are still problems with exceptions not being thrown in certain cases. I found out that if some problem happens with import errors inside included templates, Nunjucks will fail to generate the required exceptions. For example, create a template **included.njk**:

```njk
{% import 'unknown.njk' as nothing %}
String is {{ str | whatever }}
```
and the following script:

```js
const path = 'C:/users/etc/thispath';
const nunjucks = require("nunjucks");
nunjucks.configure(path);
console.log(nunjucks.renderString("{% include 'included.njk' %}", { str: 'okay' }));
```
Nunjucks silently outputs a `null` string. Some notes:

- The exception is captured correctly if the contents of the njk file are directly used as the template.
- The exception is also captured if the line with `import` is removed from the njk file.

I hope this is clear. Thanks in advance!

fdintino:
## Summary

The fix introduced in PR #1049, preventing multiple calls to the render callback, was a bit too aggressive and prevented some template errors from being surfaced when `render` or `renderString` were called synchronously.

The real bug here lies in the compiler, which should not be calling the same callback function more than once. But the way that waterfall tasks are chained makes it difficult to properly halt execution in a template. We ought to fix the underlying issue, but for now this should (hopefully) ensure that thrown errors always surface when rendering synchronously, while not causing a regression for #1029.

Closes #1272 . I suspect this also fixes #1246, but I can't confirm because the reporter of that issue hasn't provided a test case I can use.

## Checklist

* ~[*Documentation*](https://github.com/mozilla/nunjucks/tree/master/docs/) is added / updated to describe proposed change. **(N/A)**~

### DIFF
--- a/nunjucks/src/environment.js
+++ b/nunjucks/src/environment.js
@@ -476,14 +476,15 @@ class Template extends Obj {
     let didError = false;
 
     this.rootRenderFunc(this.env, context, frame, globalRuntime, (err, res) => {
-      if (didError) {
+      // TODO: this is actually a bug in the compiled template (because waterfall
+      // tasks are both not passing errors up the chain of callbacks AND are not
+      // causing a return from the top-most render function). But fixing that
+      // will require a more substantial change to the compiler.
+      if (didError && cb && typeof res !== 'undefined') {
         // prevent multiple calls to cb
-        if (cb) {
-          return;
-        } else {
-          throw err;
-        }
+        return;
       }
+
       if (err) {
         err = lib._prettifyError(this.path, this.env.opts.dev, err);
         didError = true;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,9 @@ Changelog
   Merge of [#1276](https://github.com/mozilla/nunjucks/pull/1276); fixes
   [#1198](https://github.com/mozilla/nunjucks/issues/1198). Thanks
   [ogonkov](https://github.com/ogonkovv)!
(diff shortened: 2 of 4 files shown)


nunjucks/src/environment.js
```diff
@@ -476,14 +476,15 @@ class Template extends Obj {
     let didError = false;
 
     this.rootRenderFunc(this.env, context, frame, globalRuntime, (err, res) => {
-      if (didError) {
+      // TODO: this is actually a bug in the compiled template (because waterfall
+      // tasks are both not passing errors up the chain of callbacks AND are not
+      // causing a return from the top-most render function). But fixing that
+      // will require a more substantial change to the compiler.
+      if (didError && cb && typeof res !== 'undefined') {
```
fdintino, line 483: We just that `res` is undefined, because the second call to the root render callback happens here in `compileRoot`:

https://github.com/mozilla/nunjucks/blob/63c4baf2332db08ce9c170bafe0d3696d1e293bd/nunjucks/src/compiler.js#L1129-L1133

If an error occurs inside an included template, the error will be passed up the callback chain, and the user's callback will get called with the correct error argument. But it won't cause a return from the root compiled template function, so it will still call `cb(null, this.buffer)` at the very end. `this.buffer` is always a string, even if empty, so by confirming that `res` is not undefined we can check specifically for the situation described above.

PR #1315, opened 2020-08-02
Add `map` filter

ogonkov:
## Summary

Proposed change:

Add `map` filter

## Checklist


ogonkov:
@fdintino In Python filter returns `None`, when value is "falsy", in Nunjucks it returns `undefined`, this is not works very well with filters chain, because `None` in Python is seems resolving to empty list, while in JS it just `undefined`, that couldn't be iterable. Should we return empty array in case of "falsy" value?

PR #1319, opened 2020-09-30
Add jinja-compat switch for precompile script

krassowski:
See https://stackoverflow.com/a/45440732/6646912

## Summary

Proposed change:

Add `-j/--jinja-compat` switch to precompile script allowing to turn on the precompile mode.

## Checklist

### DIFF
--- a/bin/precompile
+++ b/bin/precompile
@@ -12,7 +12,7 @@ program
 
 program
   .name('precompile')
-  .usage('[-f|--force] [-a|--filters <filters>] [-n|--name <name>] [-i|--include <regex>] [-x|--exclude <regex>] [-w|--wrapper <wrapper>] <path>')
+  .usage('[-f|--force] [-a|--filters <filters>] [-n|--name <name>] [-i|--include <regex>] [-x|--exclude <regex>] [-w|--wrapper <wrapper>] [-j|--jinja-compat] <path>')
   .arguments('<path>')
   .helpOption('-?, -h, --help', 'Display this help message')
   .option('-f, --force', 'Force compilation to continue on error')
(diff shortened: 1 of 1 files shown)


bin/precompile
```diff
@@ -48,6 +49,11 @@ if (opts.wrapper) {
   opts.wrapper = require('nunjucks-' + opts.wrapper).wrapper;
 }
 
+if (opts.jinjaCompat) {
+  var nunjucks = require('nunjucks');
+  nunjucks.installJinjaCompat();
```
ogonkov, line 54: I guess that it will likely affect all next invocations of CLI utility. I think you should uninstall it after precompilation complete.

```js
if (opts.jinjaCompat) {
  var uninstallJinjaCompat = nunjucks.installJinjaCompat();
}
precompile(...);

if (opts.jinjaCompat) {
  uninstallJinjaCompat();
}
```

krassowski: Is there `uninstallJinjaCompat`? Why would it affect subsequent CLI invocations if it only modifies in memory objects?

ogonkov: Uninstall method returned from `installJinjaCompat` invocation.

`installJinjaCompat` is a bit hack-ish, it just monkey-patch original library, so it will affect all other invocations.

krassowski: > installJinjaCompat is a bit hack-ish, it just monkey-patch original library so it will affect all other invocations

I still do not get it. We are talking about a command-line-only option and the program is going to finish execution soon and the monkey-patched methods are not going to be persisted on disk in any way, so why would we want to uninstall jinjaCompat mode? There is not going to be any subsequent invocations in the way that this command line script will cease to exist.

krassowski: Slight bump after a year. I am happy to do as instructed, just wanted to confirm.

krassowski: Done in 5e3ce1e12a249e1c805f2654f2115aeddc7b7bdf.

ogonkov (approved):

PR #1329, opened 2020-11-16
Move chokidar to peerDependencies and make it optional via peerDependenciesMeta

### LINKED ISSUE #1227: Revisit the watch option and chokidar dependency
Hello!

I was digging into what was inflating the install size of `nunjucks` and finally traced it down to the `chokidar` dependency that's included in `optionalDependencies`. At one point (in #610) it was even discussed that nunjucks just get rid of this capability entirely — which I think is the right answer overall.

(I got on this path when I was comparing [Liquid JS](https://github.com/harttle/liquidjs) to nunjucks and noticed how much tinier it was for similar capability, and it has everything to do with it bypassing the work of including a watcher.)

Even now as an optional dependency it's practically a guarantee that most people are still downloading it — it's uncommon for folks to pass in the `--no-optional` flag and AFAICT it's not documented as an installation method for nunjucks. I'd argue it'd perhaps make more sense as a peer dependency or... just not there at all. 😄 

I'd be happy to help make this a reality if there's still an appetite for it! Seems like something that could be re-created as a recipe/example of how to do it externally so people still have a path to accomplish this.

rdmurphy:
## Summary

Proposed change:

This pull request implements the discussed solution to #1227 and moves `chokidar` from `optionalDependencies` (which never did what it aspired to do) to `peerDependencies`, and uses the recently implemented `peerDependenciesMeta` field to mark it as `optional`. This means for `npm` users using a version that supports `peerDependenciesMeta`, `chokidar` will have to be separately installed. This is in particularly relevant to the newest version of `npm` (v7), which installs `peerDependencies` by default unless the package is marked as `optional`.

Closes #1227.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Unreleased
   respect `throwOnUndefined` if sort attribute is undefined.
 * Add `base` arg to
   [`int` filter](https://mozilla.github.io/nunjucks/templating.html#int).
+* Move `chokidar` to `peerDependencies` and mark it `optional` in `peerDependenciesMeta`.
 
 3.2.2 (Jul 20 2020)
 -------------------
--- a/README.md
+++ b/README.md
@@ -1,10 +1,10 @@
 # Nunjucks
 
-  [![NPM Version][npm-image]][npm-url]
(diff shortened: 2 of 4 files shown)


fdintino:
Thank you for this!

PR #1481, opened 2025-01-27
Clarify title filter behavior in Templating docs

davidjb:
## Summary

Proposed change:

This PR updates the Templating docs to correct the explanation of the behavior of `title`. Previously, the docs indicate that only the first character in a string will be upper-cased but the example (and behavior as I've just tested) is that `title` will upper-case each word in a string, not just the first character. For avoidance of doubt, https://jinja.palletsprojects.com/en/stable/templates/#jinja-filters.title clearly sets out the intended behavior.

## Checklist

### DIFF
--- a/docs/templating.md
+++ b/docs/templating.md
@@ -1809,7 +1809,7 @@ Output the sum of items in the array:
 
 ### title
 
-Make the first letter of the string uppercase:
+Make the first letter of each word in the string uppercase:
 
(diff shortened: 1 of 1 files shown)


przemub:
I've made another proposal (#1521) - sorry, I haven't seen yours. That one also specifies that the remaining letters will be turned lowercase.

PR #1534, opened 2026-07-23
fix: handle non-string input in indent filter

### LINKED ISSUE #1513: Indent filter crashes when passed unexpected input
I've found that the `indent` filter will crash Nunjucks if not passed a string. 

Minimal test case:
```nunjucks
{% set testArray = ['one', 'two'] %}

{# Crashes #}
{{ testArray | indent(8) }}
```

The error returned is `str.split is not a function`

## Why I think this is a bug
My general expectation with Nunjucks is that it fails gracefully. Usually if something isn't as expected it displays nothing and continues to render.

## What I might have expected

I think I would have expected it to do one of:
* don't do anything - return original input
* don't do anything - return null / false / undefined
* cast to string and then indent

---

This isn't an issue for my project as I was only temporarily rendering it to check progress as I go. Many of our components use `indent` filter behind the scenes. I'll be updating the nunjucks to iterate over the array for display anyway.

Hashim1999164:
## Summary
- Update the `indent` filter so it no longer throws when the input is not a string (for example an array or object).
- Numbers are coerced to strings, consistent with the `replace` filter.
- Other non-string values are returned unchanged.

Fixes #1513

## Test plan
- [x] `npx mocha -R spec tests/filters.js --grep indent`


lbesecker195 (approved):
Verified: the `replace` filter coerces a numeric `str` with `'' + str` and returns any other non-string, non-SafeString input unchanged, so `indent` now follows the same convention instead of throwing on `.split`. Tests cover both paths. LGTM.

